PR #2, opened 2019-03-22
feat(error): Parse error message

ehdsouza:
Parse error message in following order:
```
errors[0].message
error
message
```


mediumTaj:
👍 looks good!

PR #9, opened 2019-05-07
feat(IAM): Allow IAM client id/secret to be configured when instantia…

padamstx:
…ting a service

Fixes: https://github.ibm.com/arf/planning-sdk-squad/issues/851

This PR allows the IAM client id and secret values to be specified when instantiating a BaseService.  Those values are then propagated to the TokenManager struct where they are used to perform basic auth when interacting with the IAM token service to request or refresh an IAM access token.

There is a separate PR in the Go code generator to allow the IAM client id and secret values to be specified when instantiating a generated service, and those values are then passed on to the BaseService ctor in that case.

### DIFF
--- a/core/token_manager.go
+++ b/core/token_manager.go
@@ -26,9 +26,10 @@ import (
 
 // for handling token management
 const (
-	DefaultIAMURL            = "https://iam.bluemix.net/identity/token"
+	DefaultIAMURL            = "https://iam.cloud.ibm.com/identity/token"
+    DefaultIAMClientId       = "bx"
+    DefaultIAMClientSecret   = "bx"
 	DefaultContentType       = "application/x-www-form-urlencoded"
-	DefaultAuthorization     = "Basic Yng6Yng="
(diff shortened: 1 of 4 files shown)


core/token_manager.go
```diff
@@ -48,27 +49,42 @@ type TokenManager struct {
...
+func NewTokenManager(iamAPIkey string, iamURL string, userAccessToken string,
+    iamClientId string, iamClientSecret string) (*TokenManager, error) {
 	if iamURL == "" {
 		iamURL = DefaultIAMURL
 	}
+	
+	if iamClientId == "" && iamClientSecret == "" {
+	    iamClientId = DefaultIAMClientId
+	    iamClientSecret = DefaultIAMClientSecret
+	} else if iamClientId != "" && iamClientSecret != "" {
```
ehdsouza: Do we need this?

padamstx: yes because if neither of those values are specified, then we want to use the default values (bx/bx) for basic auth.
If both are specified, then we want to use them.
If only 1 is specified, then it's an error... both the id and secret need to be specified together or neither of them should be specified.

PR #16, opened 2019-08-04
feat(authentication): Add support for new authentication API

padamstx:
Fixes: https://github.ibm.com/arf/planning-sdk-squad/issues/972

This PR contains changes related to our strategic authentication design.  The various authentication mechanisms have been re-organized such that each auth type has a "Config" structure that is used to configure it, along with an "Authenticator" struct which contains the actual authenticator.
The BaseService will now simply call the configured authenticator's "Authenticate()" method to have the appropriate header added to the outbound request.


core/auth_common.go
```diff
@@ -0,0 +1,77 @@
...
+	"fmt"
+	"net/http"
+)
+
+const (
+	// Supported authentication types.
+	AUTHTYPE_BASIC  = "basic"
+	AUTHTYPE_NOAUTH = "noauth"
+	AUTHTYPE_IAM    = "iam"
+	AUTHTYPE_ICP4D  = "icp4d"
```
germanattanasio: CP4D

padamstx: So... we've officially found a new name for "ICP4D"?  I thought we were going to do that in our meeting today.   My plan here was to use the same name we do everywhere else until we decide on the new official name.

padamstx:
Since the design of the new authentication API is still in flux, I'll hold off on this PR.  When the design is complete, I'll update this PR accordingly...

PR #20, opened 2019-09-05
feat: return more information in the DetailedResponse

padamstx:
Fixes: https://github.ibm.com/arf/planning-sdk-squad/issues/999

This PR enhances the Go core's ability to return as much information as possible from the operation response in the DetailedResponse structure.
Highlights of changes:
1. In the case of a successful invocation:
    * if the response body is not JSON, we'll return it as a byte array in the new `DetailedResponse.RawResult` field.
2. In the case of an unsuccessful invocation:
    * (as before) we'll obtain an error message from the response body and return it in an error object.
    * (new) we'll decode the entire response body (presumably containing an error structure) as a map and return it in the `DetailedResponse.Result` field.
3. Added new DetailedResponse methods to retrieve the RawResult and Result (as map) values.

### DIFF
--- a/core/base_service.go
+++ b/core/base_service.go
@@ -15,6 +15,7 @@ package core
 // limitations under the License.
 
 import (
+	"bytes"
 	"crypto/tls"
 	"encoding/json"
 	"fmt"
@@ -150,15 +151,15 @@ func (service *BaseService) SetUserAgent(userAgentString string) {
 }
 
 // Request performs the HTTP request
-func (service *BaseService) Request(req *http.Request, result interface{}) (*DetailedResponse, error) {
-	// Add default headers
+func (service *BaseService) Request(req *http.Request, result interface{}) (detailedResponse *DetailedResponse, err error) {
+	// Add default headers.
 	if service.DefaultHeaders != nil {
 		for k, v := range service.DefaultHeaders {
 			req.Header.Add(k, strings.Join(v, ""))
 		}
 	}
 
-	// Check if user agent is present.
+	// Add the default User-Agent header if not already present.
 	userAgent := req.Header.Get(USER_AGENT)
 	if userAgent == "" {
 		req.Header.Add(USER_AGENT, service.UserAgent)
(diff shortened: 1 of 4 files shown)


core/base_service.go
```diff
@@ -150,60 +151,87 @@ func (service *BaseService) SetUserAgent(userAgentString string) {
...
+func (service *BaseService) Request(req *http.Request, result interface{}) (detailedResponse *DetailedResponse, err error) {
+	// Add default headers.
 	if service.DefaultHeaders != nil {
 		for k, v := range service.DefaultHeaders {
 			req.Header.Add(k, strings.Join(v, ""))
 		}
 	}
 
-	// Check if user agent is present.
+	// Add the default User-Agent header if no already present.
```
mkistler: little typo: no -> not

core/base_service.go
```diff
@@ -150,60 +151,87 @@ func (service *BaseService) SetUserAgent(userAgentString string) {
...
+			// error object containing our best guess at an error message.
+			responseMap, _ := decodeAsMap(httpResponse.Body)
+			detailedResponse.Result = responseMap
+			message = getErrorMessage(responseMap)
+		} else {
+			// Response body contains something other than JSON.
+			// Return the response body as a byte array in the RawResult field
+			// and also return the entire response body as an error message.
+			detailedResponse.RawResult, _ = ioutil.ReadAll(httpResponse.Body)
+			message = string(detailedResponse.RawResult)
```
mkistler: In Swift, we create a message from the `StatusCode`. Fortunately, there is a built-in function for that in Swift and we just use it. If there was something like that in Go, I'd recommend using it. If not, perhaps we could do something simple, like just looking at the "class" of the error (4xx vs 5xx) and setting a message from that.

core/base_service.go
```diff
@@ -150,60 +151,87 @@ func (service *BaseService) SetUserAgent(userAgentString string) {
...
 		}
+		err = fmt.Errorf(message)
+		return
 	}
 
-	if response.Result == nil && result != nil {
-		response.Result = resp.Body
+	// Operation was successful, so deserialize the response.
```
mkistler: This looks like a very significant change in the way the response is returned. Do we need to make corresponding changes to the generator (it seems like these would be needed).

core/base_service.go
padamstx: done

core/base_service.go
padamstx: done

padamstx:
Just pushed an amended commit... ready for re-review

core/base_service.go
```diff
@@ -150,60 +151,122 @@ func (service *BaseService) SetUserAgent(userAgentString string) {
...
+				detailedResponse.Result = responseMap
+				err = fmt.Errorf(getErrorMessage(responseMap, detailedResponse.StatusCode))
+				return
+			}
 		}
+
+		// For a non-JSON response or if we tripped while decoding the JSON response,
+		// just return the response body byte array in the RawResult field along with
+		// an error object that contains the generic error message for the status code.
+		detailedResponse.RawResult = responseBody
```
ehdsouza: I am not sure if we really want to put in all the parsed error in the detailed response result. Why don't we deserialize and put it in the error object? This would be more go-like according to me.

```
detailedResponse, err := service.Request(req, new(Foo))
if err != nil {
  panic(err)
}
```

padamstx: Yes, I thought about using a custom error type, but in the examples that I saw, it looked like it was more difficult for users to deal with those custom error types.
At some level, it makes sense to assign the decoded (JSON) response body (for either a successful or unsuccessful invocation) in the "Response" field.   If there was any sort of problem decoding the response body, then the raw byte array is set on RawResult.
The user can rely on the DetailedResponse.StatusCode field to know whether the operation succeeded or not and then use the Result field accordingly.

PR #21, opened 2019-09-07
Defer service URL validation

padamstx:
Fixes arf/planning-sdk-squad#1011
    
    Why:
    In order to support parameterized URLs, we need to defer the validation of
    the service URL value (i.e. myservice.Service.Options.URL) to the point at
    which it is first needed, instead of when the BaseService struct is
    constructed.
    
    What:
    This PR contains changes to support the above, plus the following:
    1) Added comments here and there to some public structs and methods.
    2) Made certain elements within BaseService private since they're not needed
    outside BaseService.
    3) Removed the Version field from the ServiceOptions struct.  Instead, the
    Go generator will be changed to define Version as a field of the generated
    service struct (e.g. AssistantV1) only when the service is detected to
    use a "version param".  The Go core has no use for the Version field as it
    is only ever referenced by the generated service code, so better to move
    the definition of the field there as well.  Plus, we only generate the
    Version-related code if we find that the API definition defines a
    "version param".

Note: There will be a corresponding PR in the Go generator as well.

### DIFF
--- a/core/base_service.go
+++ b/core/base_service.go
@@ -26,29 +26,44 @@ import (
 	"time"
 )
 
-// common constants for core
 const (
-	USER_AGENT    = "User-Agent"
-	SDK_NAME      = "ibm-go-sdk-core"
-	UNKNOWN_ERROR = "Unknown Error"
+	header_name_USER_AGENT = "User-Agent"
+	sdk_name               = "ibm-go-sdk-core"
 )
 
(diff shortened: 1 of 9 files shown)


padamstx:
FYI... this PR is stacked up behind another PR so ignore the first commit ("return more information...").   I'll rebase this PR once the first PR is merged into master.

PR #26, opened 2019-09-27
Allow multiple form parts with the same key

mkistler:
This PR modifies the Query and Form properties in the RequestBuilder to allow multiple values for the same key.

Testcase included.

### DIFF
--- a/core/request_builder_test.go
+++ b/core/request_builder_test.go
@@ -225,6 +225,45 @@ func TestURLEncodedForm(t *testing.T) {
 	assert.NotNil(t, request)
 }
 
+func TestBuildWithMultipartFormWithRepeatedKeys(t *testing.T) {
+	var str = "hello"
+	json1 := make(map[string]interface{})
+	json1["name1"] = "test name1"
+
+	json2 := make(map[string]interface{})
+	json2["name2"] = "test name2"
+
+	builder := NewRequestBuilder("POST")
+	_, err := builder.ConstructHTTPURL("test.com", nil, nil)
+	assert.Nil(t, err)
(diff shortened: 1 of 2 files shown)


core/request_builder_test.go
```diff
@@ -225,6 +225,43 @@ func TestURLEncodedForm(t *testing.T) {
...
+		AddFormData("name", "json1.json", "application/json", json1).
+		AddFormData("name", "json2.json", "application/json", json2).
+		AddFormData("hello", "", "text/plain", "Hello GO SDK").
+		AddFormData("hello", "", "text/plain", &str)
+
+	pwd, _ := os.Getwd()
+	var testFile io.ReadCloser
+	testFile, err = os.Open(pwd + "/../resources/test_file.txt")
+	assert.Nil(t, err, "Could not open file")
+	builder.AddFormData("test_file1", "test_file.txt", "application/octet-stream", testFile)
```
padamstx: Just curious why `testFile` is not `&testFile` on this line like the line after it???
I'm actually surprised that that compiles clean.

core/request_builder_test.go
```diff
@@ -225,6 +225,43 @@ func TestURLEncodedForm(t *testing.T) {
...
+	builder.AddFormData("test_file2", "test_file.txt", "application/octet-stream", &testFile)
```
padamstx: Did you intend to use duplicate part names here?

mkistler: Once again, I just copied those from the test above. I was only testing "name", but no reason not to make these duplicates since they are in the test.

PR #30, opened 2019-10-17
fix: use go module instead of dep for dependency management; use golangci-lint for linting

padamstx:
This PR changes the go core rep to use Go modules instead of the almost-obsolete "dep" tool for dependency management.
I also switched the build to use the golangci-lint tool for linting which seems to be more in vogue


go.mod
```diff
@@ -0,0 +1,17 @@
+module github.com/IBM/go-sdk-core
+
+go 1.12
+
+require (
+	github.com/davecgh/go-spew v1.1.1 // indirect
+	github.com/dgrijalva/jwt-go v3.2.0+incompatible
```
ehdsouza: Was the `incompatible` verison autogenerated?

padamstx: yes, the `go mod tidy` command must have found that one as the most recent version.
I'll check to see if there's a better version that we should be using instead.

padamstx: ok, I initially thought that the `+incompatible` part of that version tag was to be feared.  But it turns out that the go mod command adds that part automatically in certain situations.   I don't fully understand it, but I don't think it's a problem.   The most recent version of that package (as seen on github) is "v3.2.0", so we should be good.

PR #35, opened 2019-10-24
feat(BaseService): add new method ConfigureService() to BaseService struct

jorge-ibm:
Fixes: https://github.ibm.com/arf/planning-sdk-squad/issues/1131

Changes:
- moved the external configuration code outside of the constructor into it's own method.
- for compatibility, the constructor calls the new method until we move to the service factory's complete implementation.

The new method will now be invoked from the generator after the BaseService's constructor is called. I have made the method public so it is accessible from the service instance.

### DIFF
--- a/core/base_service_test.go
+++ b/core/base_service_test.go
@@ -1003,6 +1003,87 @@ func TestExtConfigFromVCAP(t *testing.T) {
 	clearTestVCAP()
 }
 
+func TestConfigureServiceFromCredFile(t *testing.T) {
+	service, err := NewBaseService(
+		&ServiceOptions{
+			Authenticator: &NoAuthAuthenticator{},
+			URL:           "bad url",
+		}, "service_1", "service_1")
+	assert.Nil(t, err)
+	assert.NotNil(t, service)
+	assert.Equal(t, "bad url", service.Options.URL)
+	assert.Nil(t, service.Client.Transport)
+
+	pwd, _ := os.Getwd()
+	credentialFilePath := path.Join(pwd, "/../resources/my-credentials.env")
+	os.Setenv("IBM_CREDENTIALS_FILE", credentialFilePath)
+
+	err = service.ConfigureService("service5")
+	assert.Nil(t, err)
+	assert.NotNil(t, service)
+	assert.Equal(t, "https://service5/api", service.Options.URL)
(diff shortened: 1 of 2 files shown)


core/base_service_test.go
```diff
@@ -1003,6 +1003,87 @@ func TestExtConfigFromVCAP(t *testing.T) {
 	clearTestVCAP()
 }
 
+func TestConfigureServiceFromCredFile(t *testing.T) {
```
padamstx: Let's discuss why you needed to add a test that explicitly calls `ConfigureService` after calling `NewBaseService` since `NewBaseService` should call `ConfigureService`.

padamstx: oh wait, you are calling ConfigureService with a different service name... clever :)
you almost fooled me, but not quite... haha

core/base_service_test.go
```diff
@@ -1003,6 +1003,87 @@ func TestExtConfigFromVCAP(t *testing.T) {
...
+
+func TestConfigureServiceFromVCAP(t *testing.T) {
+	service, err := NewBaseService(
+		&ServiceOptions{
+			Authenticator: &NoAuthAuthenticator{},
+			URL:           "bad url",
+		}, "service2", "service2")
+	assert.Nil(t, err)
+	assert.NotNil(t, service)
+  assert.Equal(t, "bad url", service.Options.URL)
```
padamstx: let's fix the indentation just to be all proper and such :)
indent should be a tab character

padamstx:
One little comment about the commit message... when using the "feat:" prefix, if you use something like
`feat(BaseService): add new method ConfigureService() to BaseService struct`
then when semantic-release builds the changelog for the new release, it will group features together under the string inside the ()'s (i.e. BaseService).

core/base_service_test.go
jorge-ibm: yeah my intention was to have this test there once we remove ConfigureService out of the constructor. Technically when we do that the test should still pass

core/base_service_test.go
jorge-ibm: it's weird, I did notice that in my PR but in my editor indentation is fine. I'll see what's going on

jorge-ibm:
thanks @padamstx , I actually didn't know that was a feature of semantic release

PR #43, opened 2019-12-02
fix: properly support text responses

padamstx:
This PR fixes an issue I found while working on improvements to Go unit test generation in the SDK generator.
Specifically, this PR adds support for receiving a text-based operation response (i.e. a response whose Content-Type header is "text/html" or "text/plain", etc.).  For text-based responses, we would expect the result to be set in the DetailedResponse.Result field as a string.   Prior to this change, text-based responses were incorrectly handled just like a byte-stream type response (i.e. the DetailedResponse.Result field would simply be set to the io.ReadCloser that wraps the bytes contained in the response body).

### DIFF
--- a/core/base_service.go
+++ b/core/base_service.go
@@ -295,9 +295,24 @@ func (service *BaseService) Request(req *http.Request, result interface{}) (deta
 			// Decode step was successful. Return the decoded response object in the Result field.
 			detailedResponse.Result = result
 			return
+		} else if IsTextMimeType(contentType) {
+			// For a "text" response, let's just read the response body into a string.
+			
+			// First, read the response body into a byte array.
+			defer httpResponse.Body.Close()
+			responseBody, readErr := ioutil.ReadAll(httpResponse.Body)
+			if readErr != nil {
+				err = fmt.Errorf("An error occurred while reading the response body: '%s'", readErr.Error())
+				return
+			}
+			
+			// Next, create a string containing the response bytes and save it in the Result field.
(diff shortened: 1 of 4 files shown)


core/base_service.go
```diff
@@ -295,9 +295,24 @@ func (service *BaseService) Request(req *http.Request, result interface{}) (deta
 			// Decode step was successful. Return the decoded response object in the Result field.
 			detailedResponse.Result = result
 			return
+		} else if IsTextMimeType(contentType) {
```
mkistler: It looks like this would be a breaking change for the Watson Go SDK.  In particular, for Personality Insights, where the `profile` operation can return a `text/csv` response.  The current SDK returns its result as an `io.ReadCloser` for this case:

https://github.com/watson-developer-cloud/go-sdk/blob/master/personalityinsightsv3/personality_insights_v3.go#L246

Another case is Language Translator V3, the `GetTranslatedDocument` method:

https://github.com/watson-developer-cloud/go-sdk/blob/2f7cd68ff50aac03d3801d3aa1d10a5b6d964124/languagetranslatorv3/language_translator_v3.go#L704

which could be called with `accept="text/plain"`.

padamstx:
Closing this as we do not want to pull in this change.   Instead, we'll ensure that for both text and non-text mimetypes, generators will use the language-specific "stream" type as the return value of the operation.  Example: InputStream (Java), io.ReadCloser (Go), etc.

PR #44, opened 2019-12-03
feat: allow JSON response body to be streamed

padamstx:
This PR will allow the generated Go code to support a scenario where the decoding of a JSON response body is bypassed with the response body being returned in the DetailedResponse.Result field as an io.ReadCloser instance.
To do this, the generated code would pass in the result of `new(io.ReadCloser)` instead of `new(Foo)` for the deserialization target result object (where Foo represents a model object).

### DIFF
--- a/core/base_service_test.go
+++ b/core/base_service_test.go
@@ -15,6 +15,8 @@ package core
 // limitations under the License.
 
 import (
+	"bytes"
+	"encoding/json"
 	"fmt"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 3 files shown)


core/base_service_test.go
```diff
@@ -143,6 +197,53 @@ func TestGoodResponseNonJSON(t *testing.T) {
...
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-type", "text/plain")
+		fmt.Fprint(w, expectedResponse)
+	}))
+	defer server.Close()
+
+	builder := NewRequestBuilder("GET")
+	_, err := builder.ConstructHTTPURL(server.URL, nil, nil)
+	assert.Nil(t, err)
+	builder.AddHeader("Content-Type", "Application/json").
```
mkistler: Did you intend to pass "application/json" here?  I suppose it is fine, just might be confusing to someone reading this code later.

padamstx: Yeah those statements were carried forward from the test method I copied from, but it turns out none of the base_service test methods need to set the content type or the version query param so I just remove it from all of the methods.

PR #45, opened 2019-12-09
refactor: created new major version 3.0.0 in v3 directory

padamstx:
This PR fixes our existing go core library so that it properly supports versioning with respect to Go modules.   Specifically, this PR introduces a new major version (v3) which will be our starting point for correctly managing versions of the go core package.   Ideally, we would have created a "v2" directory and added "/v2" to the end of the package name when we created the v2.0.0 major version, but due to inexperience with Go modules, this didn't happen.   The recommended way to fix this is to simply introduce a new major version (v3 in our case) and start doing things correctly from this point forward.

The main changes are:
- Created a new "v3" directory and moved all code and resources from the root directory to "v3".
- Updated the go.mod file to update the package name to "github.com/IBM/go-sdk-core/v3"
- Updated other files to reflect the new v3 directory as needed.


padamstx:
I've tested this change in my go-test project (I updated the import paths in the go-test code to be `github.com/IBM/go-sdk-core/v3/core` and the new version (v3.0.0-rc1 in my case) was correctly found by the Go engine.

PR #54, opened 2020-04-27
feat(unmarshal): introduce new unmarshal functions for primitives and models

padamstx:
Fixes: https://github.ibm.com/arf/planning-sdk-squad/issues/1613

This PR contains new unmarshal functions for primitive and model types in the Go core.
This will be needed in the SDK generator changes to fix the referenced issue.

The "new" approach for unmarshalling JSON values (primitives and models) in the Go generator will be to generate a single `Unmarshal<model>()` function for each model.   This generated function can then be passed as a parameter (along with the JSON input and a pointer to the unmarshal result) to the core's UnmarshalModel() function.  UnmarshalModel() will use reflection to interpret the JSON input and the result to determine the appropriate type of unmarshal operation (e.g. a single model instance, a slice of models, a map of models, or a map of model slices).

The new `UnmarshalPrimitive()` function will be invoked from within a model's generated `Unmarshal<model>()` function to unmarshal struct fields involving primitive types (e.g. a single instance of a primitive type, a slice of primitives, a map of primitives, a map of primitive slices, and a slice of maps of primitive slices).

This PR contains only the changes in the Go core.  There will be a corresponding PR for the SDK generator changes.

Note: there might be some additional tweaks needed to the code changes in this PR as I complete the corresponding SDK generator changes, but I wanted to get this PR submitted so the review process can begin.   One change that I identified as I reviewed the code before submitting this PR is that there is a common piece of code within each of the private unmarshalModelXXX functions that I can pull out and implement in a private function.   This will not change the overall behavior of those functions.

### DIFF
--- a/core/unmarshal_v2.go
+++ b/core/unmarshal_v2.go
@@ -1,0 +1,521 @@
+/**
+ * (C) Copyright IBM Corp. 2020.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package core
+
+import (
+	"encoding/json"
+	"fmt"
+	"reflect"
+)
+
+const (
+	errorPropertyInsert          = " property '%s' as"
+	errorPropertyNameMissing     = "the 'propertyName' parameter is required"
+	errorUnmarshalPrimitive      = "error unmarshalling property '%s': %s"
+	errorUnmarshalModel          = "error unmarshalling%s %s: %s"
+	errorIncorrectInputType      = "expected 'rawInput' to be a %s but was %s"
+	errorUnsupportedMapEntryType = "unsupported map entry type: %s"
+	errorUnsupportedResultType   = "unsupported 'result' type: %s"
+)
+
+//
+// UnmarshalPrimitive retrieves the specified property from 'rawInput',
+// then unmarshals the resulting value into 'result'.
+//
+// This function will typically be invoked from within a model's generated "Unmarshal<model>()" method to unmarshal
+// a struct field (model property) involving a primitive type (a scalar value, a slice, a map of the primitive, etc.).
+// In this context, "primitive" refers to a type other than a user-defined model type within an OpenAPI definition.
+// The 'rawInput' parameter is expected to be a map that contains an instance of a user-defined model type.
+//
+// Parameters:
+//
+// rawInput: the unmarshal input source in the form of a map[string]json.RawMessage
+// The value 'rawInput[propertyName]' must be a json.RawMessage that contains an instance of the type inferred
+// from the value passed in as 'result'.
+//
+// propertyName: the name of the property (map entry) to retrieve from 'rawInput'. This entry's value will
+// be used as the unmarshal input source.
+//
+// result: a pointer to the unmarshal destination. This could be any of the following:
+//   - **<primitive-type>
+//   - *[]<primitive-type>
+//   - *map[string]<primitive-type>
+//   - *map[string][]<primitive-type>
+//   - *[]map[string]<primitive-type>
+//   - *[]map[string][]<primitive-type>
+//
+// Where <primitive-type> could be any of the following:
+//   - string, bool, []byte, int64, float32, float64, strfmt.Date, strfmt.DateTime,
+//     strfmt.UUID, interface{}, or map[string]interface{}.
+//
+// Example:
+// type MyStruct struct {
+//     Field1 *string,
+//     Field2 map[string]int64
+// }
+// myStruct := new(MyStruct)
+// jsonString := `{ "field1": "value1", "field2": {"foo": 44, "bar": 74}}`
+// var rawMessageMap map[string]json.RawMessage
+// var err error
+// err = UnmarshalPrimitive(rawMessageMap, "field1", &myStruct.Field1)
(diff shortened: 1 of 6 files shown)


padamstx:
@smikulcik Hi Simon, I wanted to tag you as a reviewer on this PR, but I see you're not in the required team within the github.com/IBM org.    Please feel free to take a look and provide feedback one way or another :)

core/unmarshal_v2.go
```diff
@@ -0,0 +1,521 @@
...
+			// a map[string]RawMessage that we can pass to 'unmarshaller'.
+			rawMap = make(map[string]json.RawMessage)
+			err = json.Unmarshal(rawMsg, &rawMap)
+			if err != nil {
+				err = fmt.Errorf(errorUnmarshalModel, propInsert(propertyName),
+					reflect.TypeOf(result).Elem().Elem().String(), err.Error())
+				return
+			}
+		} else {
+			rawMap = nil
```
jorge-ibm, line 235: if ``rawMap`` is nil this method will return no error, but what is the expected unmarshal result? Will ``result`` still be an empty interface?

core/unmarshal_v2.go
```diff
@@ -0,0 +1,521 @@
...
+			rawMap = make(map[string]json.RawMessage)
+			err = json.Unmarshal(rawMsg, &rawMap)
+			if err != nil {
+				err = fmt.Errorf(errorUnmarshalModel, propInsert(propertyName),
+					reflect.TypeOf(result).Elem().String(), err.Error())
+				return
+			}
+		}
+	} else {
+		foundIt = true
```
jorge-ibm, line 395: Suggestion: set ``foundIt`` to ``true`` when you first initialize it so you can remove this extra condition. If ``rawMap`` doesn't have the property then it'll be false, otherwise it'll still be set to true

core/unmarshal_v2.go
padamstx: In the case where the specified property is not found in the input map, we want to essentially return a nil model pointer.
That's why we initialize te result to nil, then do the actual unmarshalling only if rawMap is not nil.

core/unmarshal_v2_primitives_test.go
```diff
@@ -0,0 +1,1386 @@
...
+	assert.Nil(t, err)
+	assert.Nil(t, model.PropMapSlice)
+
+	err = UnmarshalPrimitive(rawMap, "null_prop", &model.PropSliceMapSlice)
+	assert.Nil(t, err)
+	assert.Nil(t, model.PropSliceMapSlice)
+
+	// Negative tests
+	model.Prop = nil
+	err = UnmarshalPrimitive(rawMap, "bad_type", &model.Prop)
```
mkistler, line 1258: This seems wrong to me.  Type "Any" should allow any type -- how can we have a bad type?

core/unmarshal_v2_primitives_test.go
```diff
@@ -0,0 +1,1386 @@
...
+		"prop_slice": ["%u1", "%u2"],
+		"prop_map": { "key1": "%u1", "key2": "%u2" },
+		"prop_slice_map": { "key1": ["%u1", "%u2"], "key2": ["%u3", "%u4"] },
+		"prop_map_slice": [{"key1": "%u1"}, {"key2": "%u2"}],
+		"prop_slice_map_slice": [{"key1": ["%u1"]}, {"key2": ["%u2", "%u3", "%u4"]} ],
+		
+		"bad_type":  true,
+		"not_a_slice": false,
+		"bad_slice_type": [38, 26],
+		"null_prop": null
```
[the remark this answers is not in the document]
padamstx: It turns out that "" and "not a real uuid" are not treated as errors by the strfmt.UUID unmarshalling code.   Not surprising due to the fact that UUID is really just an alias for "string":
```
// UUID represents a uuid string format
//
// swagger:strfmt uuid
type UUID string
```

core/unmarshal_v2_primitives_test.go
```diff
@@ -0,0 +1,1386 @@
...
+		"prop_slice": ["%d1", "%d2"],
+		"prop_map": { "key1": "%d1", "key2": "%d2" },
+		"prop_slice_map": { "key1": ["%d1", "%d2"], "key2": ["%d3", "%d4"] },
+		"prop_map_slice": [{"key1": "%d1"}, {"key2": "%d2"}],
+		"prop_slice_map_slice": [{"key1": ["%d1"]}, {"key2": ["%d2", "%d3", "%d4"]} ],
+		
+		"bad_type":  true,
+		"not_a_slice": false,
+		"bad_slice_type": [38, 26],
+		"null_prop": null
```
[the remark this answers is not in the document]
padamstx: it turns out that "" will be unmarshalled by the strfmt.DateTime unmarshalling code to produce "1970-01-01T00:00:00.000Z", rather than produce an error.

core/unmarshal_v2_primitives_test.go
padamstx: Come on Mike, I was using alliteration in my testcase development :)  lol

core/unmarshal_v2_primitives_test.go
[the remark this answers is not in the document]
mkistler: Ouch!

core/unmarshal_v2_primitives_test.go
mkistler: hahahaha

padamstx:
I just pushed a new commit to this PR that includes the following:
- Addresses review comments
- Adds a bunch of additional tests, particularly negative tests
- While adding more tests, I realized that I needed to add tests involving models with a descriminator, and that required some minor changes to the existing function to accomodate both `**Foo` and `*FooIntf` as model result types that need to be supported.

I think (I hope) that this PR is now 100% complete and ready for the corresponding Generator changes to commence (famous last words, right?).

padamstx:
While working through the Go generator changes, I found even more opportunities to improve the Go core :)
I just pushed a commit that enhances the BaseService.Request() method so that it can directly return various types of response body data via the 'result' parameter (and the DetailedResponse.Result field as well).   These changes allow us to reduce the amount of code that we generate immediately after the call to BaseService.Request() within an operation method.

PR #55, opened 2020-04-27
fix: Pace requests to token server for new auth tokens

jorge-ibm:
This PR updates the design for token refresh by decoupling the refresh logic from active requests to be consistent with the other cores.

ref: https://github.ibm.com/arf/planning-sdk-squad/issues/1550


core/cp4d_authenticator.go
```diff
@@ -61,8 +62,14 @@ type CloudPakForDataAuthenticator struct {
 
 	// The cached token and expiration time.
 	tokenData *cp4dTokenData
+
+	// The refresh time for the current cached token.
+	tokenRefreshTime int64
 }
 
+var cp4dMuOne sync.Mutex
```
jorge-ibm, line 70: I decided to use locks instead of goroutines for achieving the synchronous design due to the simplicity. I created two mutexes because I couldn't find a straight forward answer on whether or not I can use the same lock in two places at once. My guess is no, so I created two of them to be safe.

core/cp4d_authenticator.go
```diff
@@ -144,24 +151,80 @@ func (authenticator *CloudPakForDataAuthenticator) Authenticate(request *http.Re
...
 		if err != nil {
 			return "", err
 		}
-
-		authenticator.tokenData, err = newCp4dTokenData(tokenResponse)
-		if err != nil {
-			return "", err
+	} else if authenticator.tokenData != nil && authenticator.needsRefresh() {
+		// If refresh needed, kick off a go routine in the background to get a new token
+		ch := make(chan error)
```
jorge-ibm, line 165: I created a goroutine to achieve the "multi-threading" we want. The select case below it should ensure that no other routine sits around and waits for the newly spawned routine to finish

core/cp4d_authenticator.go
```diff
@@ -144,24 +151,80 @@ func (authenticator *CloudPakForDataAuthenticator) Authenticate(request *http.Re
...
+		return nil
+	}
+
+	err := authenticator.getTokenData()
+	if err != nil {
+		return err
+	}
+
+	// If the refresh time isn't set, then set it
+	if authenticator.tokenRefreshTime == 0 && &authenticator.tokenData.ExpiresIn != nil && &authenticator.tokenData.Expiration != nil {
```
jorge-ibm, line 203: @mkistler @padamstx - now that I'm looking at this again, should I *always* be setting the correct refresh time when we request a new token? Currently, it will only be set when you request the first token, and when you refresh the token where we push it forward 1 min.. I think this will result in more token requests than we'd like

core/cp4d_authenticator.go
```diff
@@ -144,24 +151,80 @@ func (authenticator *CloudPakForDataAuthenticator) Authenticate(request *http.Re
...
+		ch := make(chan error)
+		go func() {
+			ch <- authenticator.getTokenData()
+		}()
+		select {
+		case err := <-ch:
+			if err != nil {
+				return "", err
+			}
+		default:
```
mkistler, line 174: Don't you need to update `tokenRefreshTime` here?  Otherwise you'll continually refresh the token.

core/cp4d_authenticator.go
```diff
@@ -144,24 +151,80 @@ func (authenticator *CloudPakForDataAuthenticator) Authenticate(request *http.Re
...
+// and set the tokenRefreshTime
+func (authenticator *CloudPakForDataAuthenticator) synchronizedRequestToken() error {
+	cp4dMuOne.Lock()
+	defer cp4dMuOne.Unlock()
+	// if cached token is still valid, then just continue to use it
+	if authenticator.tokenData != nil && authenticator.tokenData.isTokenValid() {
+		return nil
+	}
+
+	err := authenticator.getTokenData()
```
mkistler, line 197: I think it is very dangerous to perform a network call while holding a mutex.  We should find some other solution.

core/cp4d_authenticator.go
jorge-ibm: Yes I think I do need to re-calculate the refresh time when we get a new token. Not sure why I kept that out originally.. I think I was following the java core too much

core/cp4d_authenticator.go
mkistler: Yes ... you should be updating tokenRefreshTime whenever you get a new token.

core/cp4d_authenticator.go
mkistler: Thinking about this further ... I suppose the other languages must do something similar.  Hmmm.

padamstx: In the Java core, we do this inside the TokenRequestBasedAuthenticator's getToken() method:
- if we have no token at all, or the current token has expired, then we invoke a blocking request (synchronized) to the token server
- otherwise we have a valid token and we use that as-is, but if the token is within the refresh window (last 20% of lifetime I think), then we ALSO start a refresh in a separate thread.

Not sure what else we can do to improve on that

core/cp4d_authenticator.go
padamstx: In java, we don't pre-compute the refresh time... the Token object's "needsRefresh()" method will determine whether or not the token needs to be refreshed on the fly.

core/cp4d_authenticator.go
jorge-ibm: @mkistler - I think the only other thing I can think of to use in place of the mutex is a blocking go routine.. Though I remember reading that under hood it still uses a mutex. I'll try to find that text again

core/cp4d_authenticator.go
jorge-ibm: I realized that as I created the PR 🙂

core/cp4d_authenticator.go
mkistler: My concern here is ... what happens if the _one_ token request we make "gets lost" -- not unheard of for a network request.  Does the client-side time out after a while and simply fail the request ... allowing another to try the request again?  If that's the case then this is probably fine.

jorge-ibm:
@mkistler @padamstx - I've pushed an amended commit that adds the missing calculation of the refreshTime everytime we get a new token. I've also moved back the property ``refreshTime`` to be a part of the ``iamTokenData`` and ``cp4dTokenData`` structs, respectfully. 

As far as the concerns over using locks for network calls, it looks like by default we set a client side timeout of 30 seconds.  I've added two new tests to simulate a client side timeout, ``TestIamGetTokenTimeoutError`` and a test to simulate a 504 from the server, ``TestIamGetTokenServerError``.

Note: the client timeout doesn't cover the time spent **sending** a request rather the time spent reading/receiving.

core/cp4d_authenticator.go
padamstx: whatever happens in the Java core would also happen here, so yeah I guess a timeout would occur.   If there is some other more fault-tolerant pattern we should be using in the Node, Java, Go cores (at a minimum) to avoid this, then we'll need to put our thinking caps on and come up with a better design.

mkistler: Agreed.  I was being overly cautious.  The timeout will cover us so this should be fine.

core/cp4d_authenticator.go
```diff
@@ -63,6 +64,9 @@ type CloudPakForDataAuthenticator struct {
 	tokenData *cp4dTokenData
 }
 
+var cp4dMuOne sync.Mutex
+var cp4dMuTwo sync.Mutex
```
mkistler, line 68: I think it would be better to use meaningful names for these mutexes -- to be clear on what they lock/control.

I recommend:
- `cp4dMuOne` -> `cp4dRequestTokenMutex`
- `cp4dMuTwo` -> `cp4dNeedsRefreshMutex`

core/iam_authenticator.go
```diff
@@ -77,10 +78,12 @@ type IamAuthenticator struct {
 	tokenData *iamTokenData
 }
 
+var iamMuOne sync.Mutex
+var iamMuTwo sync.Mutex
```
mkistler, line 82: Here too ... please use meaningful names for these mutexes.

core/iam_authenticator.go
```diff
@@ -167,24 +170,74 @@ func (this IamAuthenticator) Validate() error {
...
+		case err := <-ch:
+			if err != nil {
+				return "", err
+			}
+		default:
 		}
 	}
 
+	// return an error if the access token is not valid or was not fetched
+	if authenticator.tokenData == nil || authenticator.tokenData.AccessToken == "" {
```
mkistler, line 198: It's curious that here you check `authenticator.tokenData.AccessToken == ""` but above you check `!authenticator.tokenData.isTokenValid()`.  Why the difference?  Can't we make them both the same?

core/iam_authenticator.go
```diff
@@ -167,24 +170,74 @@ func (this IamAuthenticator) Validate() error {
...
+	if authenticator.tokenData != nil && authenticator.tokenData.isTokenValid() {
+		return nil
+	}
+
+	err := authenticator.getTokenData()
+	if err != nil {
+		return err
+	}
+
+	return nil
```
mkistler, line 221: Couldn't the above sequence of 6 lines be expressed simply as:
```
    return authenticator.getTokenData()
```

core/iam_authenticator.go
jorge-ibm: Just an oversight on my end. I'll go ahead and add that

jorge-ibm: So @mkistler I went ahead and tried to make this change, though it's important to note``isTokenValid`` also checks to see if the token is expired. I think it's a good extra check to have, though the issue is that for cp4d's unit tests, the expiration time is calculated using some jwt token, and not by a mock value from actual token response. The token used in those tests is expired, so all the tests using it would fail.. Getting the unit tests to work would require either a new token that would eventually also expire, changing the ``GetCurrentTime`` method to accept a mock clock, or changing how we set and calculate the expired time for a token. I just went ahead and reverted this change.

PR #61, opened 2020-06-10
Correctly handle nil pointer interfaces in RequestBuilder.SetBodyContent

stevenxie:
`RequestBuilder.SetBodyContent` currently does not correctly handle the case when its `interface{}` arguments are nil pointers. 

In Go, a `interface{}(*string)(nil)` is not the same as `nil`. This is because `interface{}`s are essentially a box containing an underlying type and value; there are cases where this method is given arguments where the box contains an empty pointer, but the box itself is not `nil`, thereby causing the wrong body contents to be set.

I discovered this bug when trying (and failing) to use plain text with IBM's Tone Analysis API and always getting an empty reply. This is because `SetBodyContent` always encodes with the (nil) `jsonContent` instead of the (non-nil) plaintext body.

See the following delve debug logs for more information:
image


padamstx:
@stevenxie Thanks for this PR.   I apologize for not seeing it sooner but will get it merged in soon.

PR #62, opened 2020-07-13
Add test involving nil dynamic property values

padamstx:
This PR simply adds a new test which verifies that a nil dynamic property value will be correctly serialized as a JSON "null" value.

In addition to the new test, I also needed to fix an unrelated linter error found in DetailedResponse.String().

### DIFF
--- a/v4/core/detailed_response.go
+++ b/v4/core/detailed_response.go
@@ -91,5 +91,5 @@ func (response *DetailedResponse) String() string {
 	if err == nil {
 		return fmt.Sprintf("%+v\n", string(output))
 	}
-	return fmt.Sprintf("Response")
+	return fmt.Sprintf("Error marshalling DetailedResponse instance: %s", err.Error())
 }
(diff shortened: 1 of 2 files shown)


v4/core/detailed_response.go
```diff
@@ -91,5 +91,5 @@ func (response *DetailedResponse) String() string {
 	if err == nil {
 		return fmt.Sprintf("%+v\n", string(output))
 	}
-	return fmt.Sprintf("Response")
```
padamstx, line 94: I saw a linter error on this statement due to the unnecessary use of fmt.Sprintf().   This is the first go core PR I've worked on after upgrading my laptop to Fedora 32 (the latest) which uses go version 1.14.3.   Prior to this, I was running Fedora 30, which uses go 1.12.x.   I'm guessing that the upgrade to the latest Go engine is why I'm now seeing this linter error, when it wasn't seen before.   Anyway, I fixed it :)

PR #66, opened 2020-08-10
feat: add detailed error response to iam/cp4d authenticators

jorge-ibm:
This PR creates a `DetailedResponse` instance for both the IAM/CP4D authenticators in the case where the api call to either servers fails (returns a status code we consider a failure). This `DetailedResponse` instance contains the response status code, headers, and the raw byte result.

I've created a `AuthenticationError `struct that contains an instance of `DetailedResponse` along with the `err` field which contains the actual error object. Whenever we expect a detailed response to be returned from the ``Authenticate`` method due to an error, we attempt to cast the returned error into  `AuthenticationError` type. If successful, the detailed response is included in the returned value for the Base Service's ``Request`` method.

ref: https://github.ibm.com/arf/planning-sdk-squad/issues/2030

### DIFF
--- a/v4/core/authenticator.go
+++ b/v4/core/authenticator.go
@@ -21,6 +21,16 @@ import (
 // Authenticator describes the set of methods implemented by each authenticator.
 type Authenticator interface {
 	AuthenticationType() string
-	Authenticate(*http.Request) error
+	Authenticate(*http.Request) *AuthenticationError
 	Validate() error
+}
+
+// AuthenticationError describes the error returned when authentication fails
+type AuthenticationError struct {
+	Response *DetailedResponse
+	err      error
+}
+
+func (e *AuthenticationError) Error() string {
+	return e.err.Error()
 }
--- a/v4/core/authenticator.go
+++ b/v4/core/authenticator.go
@@ -21,7 +21,7 @@ import (
 // Authenticator describes the set of methods implemented by each authenticator.
 type Authenticator interface {
 	AuthenticationType() string
-	Authenticate(*http.Request) *AuthenticationError
+	Authenticate(*http.Request) error
 	Validate() error
 }
 
--- a/v4/core/base_service.go
+++ b/v4/core/base_service.go
@@ -221,8 +221,14 @@ func (service *BaseService) Request(req *http.Request, result interface{}) (deta
 
 	authError := service.Options.Authenticator.Authenticate(req)
 	if authError != nil {
-		err = fmt.Errorf(ERRORMSG_AUTHENTICATE_ERROR, authError.Error())
-		return authError.Response, err
+		castErr, ok := authError.(*AuthenticationError)
+		if !ok {
+			err = fmt.Errorf(ERRORMSG_CAST_ERROR, err)
+			return
+		}
+		detailedResponse = castErr.Response
+		err = fmt.Errorf(ERRORMSG_AUTHENTICATE_ERROR, castErr.Error())
+		return
 	}
 
 	// Invoke the request.
--- a/v4/core/base_service.go
+++ b/v4/core/base_service.go
@@ -219,10 +219,10 @@ func (service *BaseService) Request(req *http.Request, result interface{}) (deta
 		return
 	}
 
-	err = service.Options.Authenticator.Authenticate(req)
-	if err != nil {
-		err = fmt.Errorf(ERRORMSG_AUTHENTICATE_ERROR, err.Error())
-		return
+	authError := service.Options.Authenticator.Authenticate(req)
+	if authError != nil {
+		err = fmt.Errorf(ERRORMSG_AUTHENTICATE_ERROR, authError.Error())
+		return authError.Response, err
 	}
 
 	// Invoke the request.
--- a/v4/core/base_service.go
+++ b/v4/core/base_service.go
@@ -221,13 +221,11 @@ func (service *BaseService) Request(req *http.Request, result interface{}) (deta
 
 	authError := service.Options.Authenticator.Authenticate(req)
 	if authError != nil {
+		err = fmt.Errorf(ERRORMSG_AUTHENTICATE_ERROR, authError.Error())
 		castErr, ok := authError.(*AuthenticationError)
-		if !ok {
-			err = fmt.Errorf(ERRORMSG_CAST_ERROR, err)
-			return
+		if ok {
+			detailedResponse = castErr.Response
 		}
-		detailedResponse = castErr.Response
-		err = fmt.Errorf(ERRORMSG_AUTHENTICATE_ERROR, castErr.Error())
 		return
 	}
 
--- a/v4/core/base_service_test.go
+++ b/v4/core/base_service_test.go
@@ -144,7 +144,7 @@ func TestGoodResponseJSONExtraFields(t *testing.T) {
 		Authenticator: &NoAuthAuthenticator{},
 	}
 	service, _ := NewBaseService(options)
-	
+
 	var foo *Foo
 	detailedResponse, _ := service.Request(req, &foo)
 	result, ok := detailedResponse.Result.(*Foo)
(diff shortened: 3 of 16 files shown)


jorge-ibm:
After some discussion with @padamstx, there's some needed refactoring on my set of changes. I will re-request reviews once those changes are done.

mkistler:
This looks like it will be a breaking change, since it changes an external interface.  If that's correct, we just need to make sure we include the right markers in the commit message so that semantic release does the right thing.

padamstx:
> This looks like it will be a breaking change, since it changes an external interface. If that's correct, we just need to make sure we include the right markers in the commit message so that semantic release does the right thing.

Agree... I hadn't considered this angle yet, but we do in fact have code other than the Go core itself that explicitly calls the `Authenticate()` method, so we'll need to package this as a new major version.

@jorge-ibm To do this, we'll need to do two things:
1. Change the commit message to include the `BREAKING CHANGE` marker in the body of the commit message.
2. We'll need to rename the top-level `v4` directory to be `v5` and modify `.bumpversion.cfg` to reflect the new directory name

Then once this change has been merged in, we'll need to change the generator to use "v5" in the import path used for the core.
Let's discuss so we can coordinate this.

v4/core/authenticator.go
```diff
@@ -21,6 +21,16 @@ import (
 // Authenticator describes the set of methods implemented by each authenticator.
 type Authenticator interface {
 	AuthenticationType() string
-	Authenticate(*http.Request) error
+	Authenticate(*http.Request) *AuthenticationError
```
padamstx, line 24: I think we can revert this back to `error` as the return type in order to maintain compatibility in the API.   This will mean that where we call Authenticate() in BaseService.Request(), you'll need to adjust that code to cast the "error" to an "AuthenticationError" when trying to retrieve the DetailedResponse field.   I hadn't thought of this "trick" before, but should have :)

v4/core/base_service.go
```diff
@@ -219,10 +219,10 @@ func (service *BaseService) Request(req *http.Request, result interface{}) (deta
 		return
 	}
 
-	err = service.Options.Authenticator.Authenticate(req)
-	if err != nil {
-		err = fmt.Errorf(ERRORMSG_AUTHENTICATE_ERROR, err.Error())
-		return
+	authError := service.Options.Authenticator.Authenticate(req)
+	if authError != nil {
+		err = fmt.Errorf(ERRORMSG_AUTHENTICATE_ERROR, authError.Error())
```
padamstx, line 224: After initializing "err" on this line, you'd need to do a type assertion to cast "authErr" to be an AuthenticationError instance (we'll use `castErr` in this example), then set `detailedResponse = castErr.Response`.

v4/core/base_service.go
```diff
@@ -219,10 +219,10 @@ func (service *BaseService) Request(req *http.Request, result interface{}) (deta
...
+		return authError.Response, err
```
padamstx, line 225: We just need to make sure that the `detailedResponse` and `err` named return values are assigned their correct values, then just do a naked return, like before.
```suggestion
		return
```

v4/core/base_service_test.go
```diff
@@ -887,11 +887,57 @@ func TestIAMFailure(t *testing.T) {
...
 
 	var foo *Foo
-	_, err = service.Request(req, &foo)
+	detailedResponse, err := service.Request(req, &foo)
 	assert.NotNil(t, err)
+	assert.NotNil(t, detailedResponse)
+	assert.NotNil(t, detailedResponse.GetHeaders())
+	assert.NotNil(t, detailedResponse.GetRawResult())
+	statusCode := detailedResponse.GetStatusCode()
+	assert.NotNil(t, statusCode)
```
padamstx, line 896: statusCode is an int, so you can't really check it against `nil`.

v4/core/base_service_test.go
```diff
@@ -887,11 +887,57 @@ func TestIAMFailure(t *testing.T) {
...
+	assert.NotNil(t, service.Options.Authenticator)
+
+	var foo *Foo
+	detailedResponse, err := service.Request(req, &foo)
+	assert.NotNil(t, err)
+	assert.NotNil(t, detailedResponse)
+	assert.NotNil(t, detailedResponse.GetRawResult())
+	statusCode := detailedResponse.GetStatusCode()
+	headers := detailedResponse.GetHeaders()
+	assert.NotNil(t, statusCode)
```
padamstx, line 934: ditto

v4/core/base_service_test.go
```diff
@@ -1052,10 +1098,58 @@ func TestCP4DFail(t *testing.T) {
...
 
 	var foo *Foo
-	_, err = service.Request(req, &foo)
+	detailedResponse, err := service.Request(req, &foo)
+	assert.NotNil(t, err)
+	assert.NotNil(t, detailedResponse)
+	assert.NotNil(t, detailedResponse.GetHeaders())
+	assert.NotNil(t, detailedResponse.GetRawResult())
+	statusCode := detailedResponse.GetStatusCode()
+	assert.NotNil(t, statusCode)
```
padamstx, line 1107: ditto

v4/core/base_service_test.go
```diff
@@ -1052,10 +1098,58 @@ func TestCP4DFail(t *testing.T) {
...
+	assert.NotNil(t, service.Options.Authenticator)
+
+	var foo *Foo
+	detailedResponse, err := service.Request(req, &foo)
+	assert.NotNil(t, err)
+	assert.NotNil(t, detailedResponse)
+	assert.NotNil(t, detailedResponse.GetRawResult())
+	statusCode := detailedResponse.GetStatusCode()
+	headers := detailedResponse.GetHeaders()
+	assert.NotNil(t, statusCode)
```
padamstx, line 1146: one more

v4/core/base_service.go
```diff
@@ -219,9 +219,15 @@ func (service *BaseService) Request(req *http.Request, result interface{}) (deta
...
 	}
 
-	err = service.Options.Authenticator.Authenticate(req)
-	if err != nil {
-		err = fmt.Errorf(ERRORMSG_AUTHENTICATE_ERROR, err.Error())
+	authError := service.Options.Authenticator.Authenticate(req)
+	if authError != nil {
+		castErr, ok := authError.(*AuthenticationError)
+		if !ok {
+			err = fmt.Errorf(ERRORMSG_CAST_ERROOR, err)
```
jorge-ibm, line 226: This technically should never happen, but just in case.

v4/core/base_service.go
```diff
@@ -219,9 +219,15 @@ func (service *BaseService) Request(req *http.Request, result interface{}) (deta
...
 	}
 
-	err = service.Options.Authenticator.Authenticate(req)
-	if err != nil {
-		err = fmt.Errorf(ERRORMSG_AUTHENTICATE_ERROR, err.Error())
+	authError := service.Options.Authenticator.Authenticate(req)
+	if authError != nil {
+		castErr, ok := authError.(*AuthenticationError)
+		if !ok {
+			err = fmt.Errorf(ERRORMSG_CAST_ERROR, err)
```
jorge-ibm, line 226: This technically should never happen, but just in case.

v4/core/cp4d_authenticator_test.go
```diff
@@ -40,6 +40,39 @@ func TestCp4dConfigErrors(t *testing.T) {
 	assert.NotNil(t, err)
 }
 
+func TestCp4dAuthenticateFail(t *testing.T) {
```
jorge-ibm, line 43: Adding this new test to test the ``error`` returned from the ``Authenticate`` method

jorge-ibm:
@padamstx @mkistler I've pushed a commit addressing the review comments along with running `go fmt ./...` to auto-format some code.

v4/core/cp4d_authenticator.go
```diff
@@ -172,9 +172,13 @@ func (authenticator *CloudPakForDataAuthenticator) getToken() (string, error) {
 		}
 	}
 
+	authError := &AuthenticationError{}
```
padamstx, line 175: I think we should initialize authError only if we're going to return an actual error.
```suggestion
```

PR #68, opened 2020-08-17
feat: add utility function to convert primitive type slices to string slices

jorge-ibm:
This PR adds a new method, ``ConvertSlice`` which converts any non-string slice to a string slice using the go ``marshal`` method, along with string manipulation. 

I chose to return an empty slice if an error occurs during conversion to reduce the generated code needed to use this method, though if we feel it's best to return an error rather than an empty slice I'm not against that change. I've added test cases to hopefully capture all the realistic scenarios this method would be used in.

ref: https://github.ibm.com/arf/planning-sdk-squad/issues/1904

### DIFF
--- a/v4/core/utils.go
+++ b/v4/core/utils.go
@@ -22,6 +22,7 @@ import (
 	"reflect"
 	"regexp"
 	"runtime"
+	"strconv"
 	"strings"
 	"time"
 
@@ -30,6 +31,7 @@ import (
 
 // Validate is a shared validator instance used to perform validation of structs.
 var Validate *validator.Validate
+var re = regexp.MustCompile(`(?s)\[(\S*)\]`)
 
 func init() {
 	Validate = validator.New()
@@ -178,4 +180,68 @@ func PrettyPrint(result interface{}, resultName string) {
 // GetCurrentTime returns the current Unix time.
 func GetCurrentTime() int64 {
 	return time.Now().Unix()
+}
+
+// ConvertSlice Marshals 'slice' to a json string, performs
+// string manipulation on the resulting string, and converts
+// the string to a '[]string'. If 'slice' is nil, not a 'slice' type,
+// or an error occurred during conversion, an error will be returned
+func ConvertSlice(slice interface{}) (s []string, err error) {
+	inputIsSlice := false
+
+	if IsNil(slice) {
+		err = fmt.Errorf(ERRORMSG_NIL_SLICE)
+		return
+	}
+
+	// Reflect on 'slice' to validate the input is in fact a slice
+	rResultType := reflect.TypeOf(slice)
+
+	switch rResultType.Kind() {
+	case reflect.Slice:
+		inputIsSlice = true
+	default:
+	}
+
+	// If it's not a slice, just return an error
+	if !inputIsSlice {
+		err = fmt.Errorf(ERRORMSG_PARAM_NOT_SLICE)
+		return
+	} else if reflect.ValueOf(slice).Len() == 0 {
+		s = []string{}
+		return
+	}
+
+	jsonBuffer, err := json.Marshal(slice)
+	if err != nil {
+		err = fmt.Errorf(ERRORMSG_MARSHAL_SLICE, err.Error())
+		return
+	}
+
+	jsonString := string(jsonBuffer)
+
+	// Use regex to convert the json string to a string slice
+	match := re.FindStringSubmatch(jsonString)
+	if match != nil && match[1] != "" {
+		newString := match[1]
+		s = strings.Split(newString, ",")
+		// For each slice element, attempt to remove any surrounding quotes
+		// added by marshaling into a json string
+		for i := range s {
+			unquotedString, unquoteErr := strconv.Unquote(s[i])
+			if unquoteErr == nil && unquotedString != "" {
+				s[i] = unquotedString
+			}
+		}
+		return
+	}
+
+	// If we returned a plain string that's not in "slice format",
+	// then attempt to just convert it to a string slice.
+	if jsonString != "" {
+		s = strings.Split(jsonString, ",")
+		return
+	}
+
+	return nil, fmt.Errorf(ERRORMSG_CONVERT_SLICE)
 }
--- a/v4/core/utils_test.go
+++ b/v4/core/utils_test.go
@@ -16,7 +16,9 @@ package core
 
 import (
 	"testing"
+	"time"
 
+	"github.com/go-openapi/strfmt"
 	"github.com/stretchr/testify/assert"
 )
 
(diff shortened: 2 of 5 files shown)


v4/core/utils.go
```diff
@@ -179,3 +179,57 @@ func PrettyPrint(result interface{}, resultName string) {
 func GetCurrentTime() int64 {
 	return time.Now().Unix()
 }
+
+// ConvertSlice Marshals 'slice' to a json string, performs
+// string manipulation on the resulting string, and converts
+// the string to a '[]string'. If 'slice' is nil, not a 'slice' type,
+// or an error occurred during conversion, an empty slice will be returned.
+func ConvertSlice(slice interface{}) (s []string) {
```
jorge-ibm, line 187: I couldn't find a way to restrict this method to only accept a parameter that is a slice, so I had to make the input an ``interface{}`` then use reflection to validate if in fact it was a slice.

v4/core/utils.go
```diff
@@ -179,3 +179,57 @@ func PrettyPrint(result interface{}, resultName string) {
...
+	if err != nil {
+		return
+	}
+
+	// Convert the json buffer to a string and remove any escaped quotes
+	jsonString := strings.Replace(string(jsonBuffer), "\"", "", -1)
+	jsonString = strings.Replace(jsonString, "\\", "", -1)
+
+	// Use regex to convert the json string to a string slice
+	re := regexp.MustCompile(`(?s)\[(\S*)\]`)
```
jorge-ibm, line 219: A simple regex that validates that the returned string is in the format: "[(any character(s))]"

v4/core/utils_test.go
```diff
@@ -192,3 +194,221 @@ func TestPointers(t *testing.T) {
...
+func TestConvertSliceString(t *testing.T) {
+	stringSlice := []string{"testString1", "testString2"}
+	expected := []string{"testString1", "testString2"}
+	convertedSlice := ConvertSlice(stringSlice)
+
+	assert.NotNil(t, convertedSlice)
+	assert.NotEmpty(t, convertedSlice)
+	assert.Equal(t, expected, convertedSlice)
+
+	stringSlice = []string{"\"testString1\"", "\"testString2\""}
```
padamstx, line 329: I think if a string within the slice actually contains double quotes, then we want those double quotes to appear in the resulting string slice that is returned.

v4/core/utils.go
padamstx, line 187: I think to be on the safe side, we should probably add an "error" return value.

v4/core/utils.go
```diff
@@ -179,3 +179,57 @@ func PrettyPrint(result interface{}, resultName string) {
...
+// ConvertSlice Marshals 'slice' to a json string, performs
+// string manipulation on the resulting string, and converts
+// the string to a '[]string'. If 'slice' is nil, not a 'slice' type,
+// or an error occurred during conversion, an empty slice will be returned.
+func ConvertSlice(slice interface{}) (s []string) {
+	s = []string{}
+	inputIsSlice := false
+
+	if IsNil(slice) {
+		return
```
padamstx, line 192: a "nil" slice should probably be considered an error

v4/core/utils.go
```diff
@@ -179,3 +179,57 @@ func PrettyPrint(result interface{}, resultName string) {
...
+	}
+
+	// Reflect on 'slice' to validate the input is in fact a slice
+	rResultType := reflect.TypeOf(slice)
+
+	switch rResultType.Kind() {
+	case reflect.Slice:
+		inputIsSlice = true
+	default:
```
padamstx, line 201: a non-slice input should be an error

v4/core/utils.go
```diff
@@ -179,3 +179,57 @@ func PrettyPrint(result interface{}, resultName string) {
...
+		return
+	}
+
+	jsonBuffer, err := json.Marshal(slice)
+	if err != nil {
+		return
+	}
+
+	// Convert the json buffer to a string and remove any escaped quotes
+	jsonString := strings.Replace(string(jsonBuffer), "\"", "", -1)
```
padamstx, line 215: I think after serializing to a JSON string, we want to retain any escaped quotes, and only remove any quotes surrounding the value of each slice element.   IOW, we need to be really careful about what we're removing so as not to remove any characters that were actually contained in the elements of the input slice.

v4/core/utils_test.go
```diff
@@ -192,3 +194,221 @@ func TestPointers(t *testing.T) {
...
+	convertedSlice = ConvertSlice(intSlice)
+
+	assert.NotNil(t, convertedSlice)
+	assert.Empty(t, convertedSlice)
+}
+
+func TestConvertSliceByteArray(t *testing.T) {
+	testString := "test string 1..."
+	byteArray := []byte(testString)
+	byteArraySlice := [][]byte{byteArray}
```
padamstx, line 265: Let's have at least 2 elements in the slice.

v4/core/utils.go
jorge-ibm: This took me a while to figure out, but I added an alternative solution that only removes the _starting_ and _ending_ surrounding quotes added by the json marshaler, if present.

v4/core/utils_test.go
jorge-ibm: Done

jorge-ibm:
@padamstx - I've changed the method to return an error, along with the other changes you requested.

v4/core/utils.go
```diff
@@ -179,3 +180,65 @@ func PrettyPrint(result interface{}, resultName string) {
...
+
+	switch rResultType.Kind() {
+	case reflect.Slice:
+		inputIsSlice = true
+	default:
+	}
+
+	// If it's not a slice or if it's an empty slice, just return
+	if !inputIsSlice || reflect.ValueOf(slice).Len() == 0 {
+		err = fmt.Errorf("the 'slice' parameter must be a non-empty slice")
```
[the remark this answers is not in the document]
padamstx: Also, my intuition tells me that we might have to treat an empty slice as valid input, which would result in us simply returning an empty string slice 's'.  This will be determined by whether or not the generated code will in fact end up calling this function for an empty slice.  Think of an optional query parameter that is a Date slice, for example.  We'll need to determine the precise code that will be generated for a scenario like that in order to know whether we could potentially pass in an empty input slice.
My suggestion is to hold off on merging this PR until after the generator changes are at least mostly finished when you'll know exactly what will be generated in those cases.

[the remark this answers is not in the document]
jorge-ibm: After looking at the generated code for go, I think for optional query params, the nil check we do before adding it to the operation's query should check to make sure that param was at least initialized before we call the new function. Either way, I've updated the function to return an empty string slice in the case the parameter is an empty slice

PR #71, opened 2020-09-08
Empty path param improvements

christiancompton:
This PR adds error handling in the core when Path parameters are the empty string.  It handles this explicitly by failing in the ConstructHTTPURL method, as well as using the annotations when validating the struct.


v4/go.sum
```diff
@@ -1,3 +1,5 @@
+github.com/IBM/go-sdk-core v1.1.0 h1:pV73lZqr9r1xKb3h08c1uNG3AphwoV5KzUzhS+pfEqY=
```
padamstx, line 1: Not sure why your environment somehow references a really OLD version of the go core, but this change should not be merged in.

padamstx (changes requested):
The code changes look fine, but I think we should hold off on merging this PR for two reasons:
1. I'm concerned with the proposed change to go.sum
2. There is another recently-opened issue that will require us to potentially re-write the function related to resolving path parameter references within a path string.

padamstx:
@christiancompton I'm going to just close this PR, as I'm going to be delivering a larger set of changes related to path params (i.e. a different way of handling them altogether) and I'll include these changes in that larger PR.

PR #72, opened 2020-09-08
feat: add new RequestBuilder.ResolveRequestURL

padamstx:
Fixes arf/planning-sdk-squad#2152

This commit adds method ResolveRequestURL to the RequestBuilder struct
as a replacement for ConstructHTTPURL().

### DIFF
--- a/v4/core/request_builder.go
+++ b/v4/core/request_builder.go
@@ -47,8 +47,9 @@ const (
 	CONTENT_TYPE            = "Content-Type"
 	FORM_URL_ENCODED_HEADER = "application/x-www-form-urlencoded"
 
-	ERRORMSG_SERVICE_URL_MISSING = "The service URL is required."
-	ERRORMSG_SERVICE_URL_INVALID = "There was an error parsing the service URL: %s"
+	ERRORMSG_SERVICE_URL_MISSING = "service URL is empty"
+	ERRORMSG_SERVICE_URL_INVALID = "error parsing service URL: %s"
+	ERRORMSG_PATH_PARAM_EMPTY    = "path parameter '%s' is empty"
 )
 
 // FormData stores information for form data.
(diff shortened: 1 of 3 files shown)


v4/core/request_builder.go
```diff
@@ -98,13 +99,78 @@ func (requestBuilder *RequestBuilder) ConstructHTTPURL(serviceURL string, pathSe
 		}
 
 		if pathParameters != nil && i < len(pathParameters) {
+			if pathParameters[i] == "" {
+				return requestBuilder, fmt.Errorf(ERRORMSG_PATH_PARAM_EMPTY, fmt.Sprintf("[%d]", i))
```
padamstx, line 103: In the old function, there's no way for us to know the name of the path parameter so I went ahead and just included its index in the "pathParameters" array within the error string.  It's certainly not ideal, but the index is at least related to the order in which the path param was defined within the operation.

v4/core/request_builder.go
```diff
@@ -47,8 +47,9 @@ const (
 	CONTENT_TYPE            = "Content-Type"
 	FORM_URL_ENCODED_HEADER = "application/x-www-form-urlencoded"
 
-	ERRORMSG_SERVICE_URL_MISSING = "The service URL is required."
-	ERRORMSG_SERVICE_URL_INVALID = "There was an error parsing the service URL: %s"
+	ERRORMSG_SERVICE_URL_MISSING = "service URL is empty"
```
padamstx, line 50: I changed these error message text strings to sort of get them more aligned with the Go style (incomplete sentences, etc.).

jorge-ibm (approved):

v4/core/request_builder.go
padamstx: > I'm trying to understand your comment above.. Did you mean to have
> 
> ```
> 				return requestBuilder, fmt.Errorf(ERRORMSG_PATH_PARAM_EMPTY, fmt.Sprintf("[%d]", pathParameters[i]))
> ```
> 
> here instead of just returning the index ?

No, in the new function, we actually have access to the name of the path param that might have a value of "", so the error message would be something like: `path parameter 'resource_id' is empty`
However, in the old function, we were not previously checking for a "" value, so when I added that check I needed to somehow identify which path parameter has the "" value.   The only thing I had access to was the index into the pathParameters array, so the error message would look something like:  `path parameter '[1]' is empty`
It's not great, but it is something :)

PR #74, opened 2020-09-16
feat(IAM Authenticator): add support for optional 'scope' property

padamstx:
The IAM /identity/token (get token)  operation supports an optional "scope"
form parameter which can be used to obtain an IAM access token
having a limited set of scopes in this it can be used.
This commit adds support for the "Scope" field within
the IamAuthenticator struct.  If set, the value of this field
is sent as the "scope" form param in the "get token"
request body when obtaining an access token.


padamstx:
@NITZANN could you please take a look at this PR and make sure it meets your needs?   Thanks

jorge-ibm (approved):
LGTM

PR #75, opened 2020-09-25
fix: dont panic in SetBodyContent when nonJSONContent is nil

dpopp07:
Fixes a bug in which not setting any body type lead to an attempt to reflect the type of a `nil` object, which caused a panic. Explicitly handles the case in which all three potential body objects are `nil`. Adds a test for this case.


padamstx (approved):
LGTM

PR #76, opened 2020-10-02
feat: add support for gzip compression of request bodies

padamstx:
Fixes arf/planning-sdk-squad#2185

This commit adds support for peforming gzip compression
of request bodies, and consists of the following:
1. Two new functions: NewGzipCompressionReader() and
NewGzipDecompressionReader().  Each function will wrap
an existing io.Reader to provide compression/decompression
using a filter pattern.

2. A new field "EnableGzipCompression" was added to the
RequestBuilder struct to indicate whether or not request
bodies should be gzip-compressed.

3. RequestBuilder.Build() was updated to honor the new flag
and (if enabled) wrap the Body field in a gzip compression filter
prior to constructing the http.Request instance.

### DIFF
--- a/v4/core/request_builder.go
+++ b/v4/core/request_builder.go
@@ -44,6 +44,7 @@ const (
 	Accept                  = "Accept"
 	APPLICATION_JSON        = "application/json"
 	CONTENT_DISPOSITION     = "Content-Disposition"
+	CONTENT_ENCODING        = "Content-Encoding"
 	CONTENT_TYPE            = "Content-Type"
 	FORM_URL_ENCODED_HEADER = "application/x-www-form-urlencoded"
 
@@ -67,6 +68,14 @@ type RequestBuilder struct {
 	Body   io.Reader
 	Query  map[string][]string
 	Form   map[string][]FormData
+
+	// EnableGzipCompression indicates whether or not request bodies
+	// should be gzip-compressed.
+	// This field has no effect on response bodies.
+	// If enabled, the Body field will be gzip-compressed and
+	// the "Content-Encoding" header will be added to the request with the
+	// value "gzip".
+	EnableGzipCompression bool
 }
 
 // NewRequestBuilder initiates a new request.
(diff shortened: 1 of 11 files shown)


v4/core/request_builder.go
```diff
@@ -305,6 +314,17 @@ func (requestBuilder *RequestBuilder) Build() (*http.Request, error) {
 		}
 	}
 
+	// If we have a request body and gzip is enabled, then wrap the body in a Gzip compression reader
+	// and add the "Content-Encoding: gzip" request header.
+	if !IsNil(requestBuilder.Body) && requestBuilder.EnableGzipCompression {
```
jorge-ibm, line 319: We should add a check to make sure the `Content-Encoding` header was not already set: https://github.com/IBM/java-sdk-core/blob/master/src/main/java/com/ibm/cloud/sdk/core/http/gzip/GzipRequestInterceptor.java#L31

jorge-ibm (changes requested):
Looks good so far but I think a few other things are missing:
- Adding code to have gzip enabled via external config
- At least 1 tests that tests enabling then disabling gzip

padamstx:
> Looks good so far but I think a few other things are missing:
> 
>     * Adding code to have gzip enabled via external config
> 
>     * At least 1 tests that tests enabling then disabling gzip

haha, thank goodness someone is paying attention... I completely whiffed on the external config requirement.  Completely forgot about it :)

v4/core/request_builder.go
padamstx: Actually, I don't think that's the logic that we want, tbh...   We should discuss.

padamstx: To expand on this a bit...  In general, HTTP headers can have multiple values.  Content-Encoding is perhaps one of those headers that could conceivably have multiple values because it's possible that the payload has had multiple "encodings" performed on it.   Perhaps our check should be that the list of values returned for the Content-Encoding header does not contain "gzip".

padamstx:
I've added support for external config properties and that also required that I add the EnableGzipCompression flag to the ServiceOptions struct defined in base_service.go (initially I was planning on just adding that to the generated service operations struct).

v4/core/request_builder.go
jorge-ibm: okay, if that needs to be updated in java then I can make that change.

jorge-ibm (approved):

PR #80, opened 2020-10-14
feat: support use of Context with RequestBuilder

### LINKED ISSUE #77: Support for context timeout/deadline for requests in underlying HTTP client
**Is your feature request related to a problem? Please describe.**
The default behavior for whenever a request is made is using `http.NewRequest` [(from here)](https://github.com/IBM/go-sdk-core/blob/master/v4/core/request_builder.go#L309) which will use the default HTTP clients timeout [(from here)](https://github.com/IBM/go-sdk-core/blob/master/v4/core/base_service.go#L86). This will block the caller and force them to wait until the request either fulfills or times out.

**Describe the solution you'd like**
Support passing in a context throughout different request types, e.g.:
```go
ctx, cancelFunc := context.WithDeadline(context.Background(), someTime)
client.GetDocumentAsStream(ctx, s.NewGetDocumentOptions(someDatabaseName, someID))
```
By supporting this, the request can be canceled so that the caller isn't blocked by the library if they have their own timeout they need to support.

padamstx:
Fixes arf/planning-sdk-squad#2230
Fixes #77

This commit adds support for setting a Context
instance on the RequestBuilder struct.  This Context instance
will then be associated with the http.Request instance constructed
by the RequestBuilder.Build() method.
The Context can be used to define a timeout/deadline for a request
or to cancel an in-flight request.

### DIFF
--- a/v4/core/request_builder.go
+++ b/v4/core/request_builder.go
@@ -16,6 +16,7 @@ package core
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -76,6 +77,10 @@ type RequestBuilder struct {
 	// the "Content-Encoding" header will be added to the request with the
 	// value "gzip".
 	EnableGzipCompression bool
+
+	// RequestContext is an optional Context instance to be associated with the
+	// http.Request that is constructed by the Build() method.
+	ctx context.Context
 }
 
 // NewRequestBuilder initiates a new request.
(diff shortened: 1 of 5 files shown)


padamstx:
@jorge-ibm @mkistler This is a replacement PR for the original which I have closed.

v4/core/request_builder.go
```diff
@@ -76,6 +77,10 @@ type RequestBuilder struct {
 	// the "Content-Encoding" header will be added to the request with the
 	// value "gzip".
 	EnableGzipCompression bool
+
+	// RequestContext is an optional Context instance to be associated with the
+	// http.Request that is constructed by the Build() method.
+	ctx context.Context
```
jorge-ibm, line 83: Do we want this field to be exportable/public ?

jorge-ibm (approved):
LGTM. Approved with 1 question

mkistler (approved):
Looks good! 👍

v4/core/request_builder.go
padamstx: Nope, I wanted to intentially "hide" it behind the "WithContext()" method

PR #81, opened 2020-10-22
fix: jwt dependency upgrade

christiancompton:
This upgrade changes the `dgrijalva/jwt-go v3.2.0` to use `form3tech-oss/jwt-go v3.2.1`, a fork which as addressed the high severity security vulnerability.

There has been a lot of conversation about this fix in https://github.com/dgrijalva/jwt-go/issues/428, but this project is not maintained by its authors and a fix to this project seems unlikely given this conversation.

### DIFF
--- a/v4/go.mod
+++ b/v4/go.mod
@@ -3,7 +3,7 @@ module github.com/IBM/go-sdk-core/v4
 go 1.12
 
 require (
-	github.com/dgrijalva/jwt-go v3.2.0+incompatible
+	github.com/form3tech-oss/jwt-go v3.2.1+incompatible
 	github.com/go-openapi/strfmt v0.19.5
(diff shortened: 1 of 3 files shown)


v4/go.mod
```diff
@@ -4,6 +4,7 @@ go 1.12
 
 require (
 	github.com/dgrijalva/jwt-go v3.2.0+incompatible
```
padamstx, line 6: we don't need to keep the original module reference, do we???

padamstx: also, if you changed go.mod to use a completely different package (i.e. form3tech-oss/jwt-go), then how did you avoid having to make source changes where the jwt-go package is being used?

christiancompton: Fixed.

padamstx (approved):
LGTM

PR #82, opened 2020-10-23
feat: introduce support for automatic retries

padamstx:
Fixes arf/planning-sdk-squad#2229

This commit includes:
- support for automatic retries of api invocations
- a new basic logger capability
- introduction of build tags in the unit tests to enable
  running various subsets of tests

In order to get a good overall view of what's going on during the tests, I suggest that reviewers pull the feature branch and then run the retry-related tests with `cd v4/core; go test -ginkgo.v -tags=retries`.  That will show the various retries being performed and the resulting status code for each request.

Re: the logger, my intent was to provide a basic logger that supports various logging levels that in turn uses the underlying `log.Logger` facility provided by the Go engine.   I define the core.Logger interface so that anyone could implement their own and supply that as the Logger instance used by the core.   This might be useful in CLI and/or Terraform contexts.


Makefile
```diff
@@ -8,10 +8,10 @@ build:
 	cd ${VDIR} && go build ./...
 
 test:
-	cd ${VDIR} && go test ./...
+	cd ${VDIR} && go test -tags=all ./...
```
padamstx, line 11: FYI... The go core unit test bucket has grown a bit over time and some of the tests (e.g. authenticator tests, "context"-related tests, etc.) are "slow" (relatively speaking).  Up to this point, I've typically just used the `-run=<regexp>` option on the `go test` command to run only a subset of the tests while doing development.   But, as part of this PR, I've introduced the use of ginkgo/gomega for the new retry-related tests since that seemed the most natural way to write them (similar to our generated unit tests, etc.), and that means that the `-run=<regexp>` option is no longer as useful as it was.   So, I went ahead and added build flags to each of the unit test files to sort of classify the sets of tests.  All _test.go files have the tag "all" so that when we run tests with `-tags=all` they all get run.    Other tags that were added are "fast", "slow", "auth", "basesvc", "log" and "retries".   To run only the "fast" tests, you'd use `go test -tags=fast`.   To run only the retry-related tests, use `go test -tags=retries`, etc.
The Makefile (used in Travis and in local sandboxes) will continue to run all the tests.

v4/core/base_service.go
```diff
@@ -16,20 +16,27 @@ package core
...
+
+	cleanhttp "github.com/hashicorp/go-cleanhttp"
+	retryablehttp "github.com/hashicorp/go-retryablehttp"
 )
 
 const (
-	header_name_USER_AGENT = "User-Agent"
-	sdk_name               = "ibm-go-sdk-core"
+	headerNameUserAgent = "User-Agent"
+	sdkName             = "ibm-go-sdk-core"
```
padamstx, line 39: Got sick of the linter warnings for these constants :)

v4/core/base_service.go
```diff
@@ -180,11 +185,18 @@ func (service *BaseService) SetHTTPClient(client *http.Client) {
 }
 
 // DisableSSLVerification skips SSL verification.
+// This function sets a new http.Client instance on the service
+// and configures it to bypass verification of server certificates
+// and host names, making the client susceptible to "man-in-the-middle"
+// attacks.  This should be used only for testing.
 func (service *BaseService) DisableSSLVerification() {
```
padamstx, line 192: Because the client being used by the BaseService might now be a retryable client, we can't just replace the Transport field to disable ssl.  Instead, we create a new default client and then modify it's Transport's tls config.
The side-effect is that by disabling SSL, a new client instance is set on the BaseService, which sounds like an abrupt change, but we were previously replacing the Transport anyway so this is not much different now.

v4/core/base_service.go
```diff
@@ -256,8 +268,25 @@ func (service *BaseService) Request(req *http.Request, result interface{}) (deta
 		return
 	}
 
-	// Invoke the request.
-	httpResponse, err := service.Client.Do(req)
+	var httpResponse *http.Response
+
+	// Try to get the retryable Client hidden inside service.Client
+	retryableClient := getRetryableHTTPClient(service.Client)
```
padamstx, line 274: If the BaseService.Client field is a "retryable" client, then it will be an instance of http.Client with an alternate transport impl which contains the entire "retryable" client instance.   This step here is to check if we in fact are using a retryable client and to return it (from the transport) if so.   We do this so we can invoke the retryable client differently from a normal client.

v4/core/base_service.go
```diff
@@ -451,3 +480,157 @@ func getErrorMessage(responseMap map[string]interface{}, statusCode int) string
...
+func (service *BaseService) EnableRetries(maxRetries int, maxRetryInterval time.Duration) {
+	client := NewRetryableHTTPClient()
+	if maxRetries > 0 {
+		client.RetryMax = maxRetries
+	}
+	if maxRetryInterval > 0 {
+		client.RetryWaitMax = maxRetryInterval
+	}
+
+	service.SetHTTPClient(client.StandardClient())
```
padamstx, line 497: The `StandardClient()` method returns an http.Client instance with the retryable client "hidden" in the transport.   This means that the BaseService.Client field will always hold an http.Client instance.   If it's a retryable client, then the transport holds the actual go-retryablehttp.Client instance.

v4/core/base_service.go
```diff
@@ -451,3 +480,157 @@ func getErrorMessage(responseMap map[string]interface{}, statusCode int) string
...
+
+func (l *httpLogger) Printf(format string, inserts ...interface{}) {
+	GetLogger().Log(LevelDebug, format, inserts...)
+}
+
+// NewRetryableHTTPClient returns a new instance of go-retryablehttp.Client
+// with a default configuration that supports Go SDK usage.
+func NewRetryableHTTPClient() *retryablehttp.Client {
+	client := retryablehttp.NewClient()
+	client.Logger = &httpLogger{}
```
padamstx, line 523: We'll use a "logger" that logs messages at "debug" level with the retryablehttp function.   This means that if the default go core logger (error level) is in use, no messages from the retryablehttp package will be logged.   If the go core logger is logging at "debug" level, then we'll see the messages.

v4/core/base_service.go
```diff
@@ -451,3 +480,157 @@ func getErrorMessage(responseMap map[string]interface{}, statusCode int) string
...
+	schemeErrorRe = regexp.MustCompile(`unsupported protocol scheme`)
+)
+
+// IBMCloudSDKRetryPolicy provides a default implementation of the CheckRetry interface
+// associated with a retryablehttp.Client.
+// This function will return true if the specified request/response should be retried.
+func IBMCloudSDKRetryPolicy(ctx context.Context, resp *http.Response, err error) (bool, error) {
+	// This logic was adapted from go-relyablehttp.ErrorPropagatedRetryPolicy().
+
+	// Do not retry on a Context-related error (Canceled or DeadlineExceeded).
```
padamstx, line 563: A timeout error would be in this category.

v4/core/base_service_retries_test.go
```diff
@@ -0,0 +1,461 @@
...
+	_, err = builder.ConstructHTTPURL(url, nil, nil)
+	Expect(err).To(BeNil())
+	return
+}
+
+var _ = Describe(`Retry scenarios`, func() {
+	var server *httptest.Server
+
+	BeforeEach(func() {
+		SetLogger(NewLogger(LevelDebug, log.New(GinkgoWriter, "", log.LstdFlags)))
```
padamstx, line 64: This step will create a new go core logger at "debug" level and use the test frameworks io.Writer instance as the output "stream".   This means that when you run the tests, you'll see the debug messages if you run with the `-ginkgo.v` option (e.g. `go test -ginkgo.v -tags=retries`).

v4/core/base_service_test.go
```diff
@@ -30,12 +32,8 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-type Foo struct {
```
padamstx, line 33: Due to the addition of the build flags, I moved common test-related definitions into a new file to be shared among the tests.

v4/core/base_service_test.go
```diff
@@ -30,12 +32,8 @@ import (
...
-	Name *string `json:"name,omitempty"`
-}
-
 // Test a normal JSON-based response.
-func TestGoodResponseJSON(t *testing.T) {
+func TestRequestGoodResponseJSON(t *testing.T) {
```
padamstx, line 36: Just modified the names of the functions so that we can use `go test -tags=all -run=TestRequest` to run just the "request"-related tests.

v4/core/marshal_nulls_test.go
```diff
@@ -1,11 +1,14 @@
...
 package core
 
 import (
 	"bytes"
 	"encoding/json"
-	"github.com/stretchr/testify/assert"
 	"reflect"
 	"testing"
+
+	"github.com/stretchr/testify/assert"
```
padamstx, line 11: I think vscode didn't like the order of the imports :)

padamstx:
cc: @smikulc please provide feedback if you want :)

v4/core/base_service.go
```diff
@@ -180,11 +185,31 @@ func (service *BaseService) SetHTTPClient(client *http.Client) {
...
+		tr.TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
 	}
-	service.Client.Transport = tr
+
+	service.SetHTTPClient(client)
+}
+
+// IsSSLDisabled returns true if and only if the service's http.Client instance
+// is configured to skip verification of server SSL certificates.
+func (service *BaseService) IsSSLDisabled() bool {
```
padamstx, line 204: While working on the corresponding generator changes (specifically the generated unit tests), I discovered that we need a more reliable way to check to see if a service has had ssl verification disabled or not, so I'm exposing this new public method on a BaseService.

jorge-ibm (commented):
This looks good. Though I pulled the branch and ran the tests locally and got the following test failure:
```go
• Failure [1.239 seconds]
Retry scenarios
/Users/jrangel/Development/devx/go-sdk-core/v4/core/base_service_retries_test.go:60
  Error scenarios
  /Users/jrangel/Development/devx/go-sdk-core/v4/core/base_service_retries_test.go:67
    Misc. errors
    /Users/jrangel/Development/devx/go-sdk-core/v4/core/base_service_retries_test.go:136
      Invalid host [It]
      /Users/jrangel/Development/devx/go-sdk-core/v4/core/base_service_retries_test.go:199

      Expected
          <string>: Get http://notahost:12345: dial tcp 23.217.138.110:12345: connect: connection refused
      to contain substring
          <string>: no such host

      /Users/jrangel/Development/devx/go-sdk-core/v4/core/base_service_retries_test.go:208
```
It looks like the travis build hasn't finished to confirm if it's just something in my sandbox.

We should also close https://github.ibm.com/arf/planning-sdk-squad/issues/2079 if we feel like this PR satisfied the requirements for adding a logger to go.

padamstx:
> This looks good. Though I pulled the branch and ran the tests locally and got the following test failure:
> 

>       Expected
>           <string>: Get http://notahost:12345: dial tcp 23.217.138.110:12345: connect: connection refused
>       to contain substring
>           <string>: no such host

Aha! It looks like I was being too specific in the error message returned for an "invalid hostname" situation.   This is likely just differences in the TCP/IP implementations on linux vs MacOS.
I'll just relax the checks a bit in the unit tests.

jorge-ibm (approved):

padamstx:
> We should also close https://github.ibm.com/arf/planning-sdk-squad/issues/2079 if we feel like this PR satisfied the requirements for adding a logger to go.

While I've added a basic logging capability to the Go core to support the retryable request processing stuff, I think we want to keep that issue open so that we can add logging steps where needed throughout the Go core code to log errors and perhaps some debug or info messages at strategic places.

mkistler (approved):
Looks good! 👍 

I really like the improvements in the tests.

I left a few minor comments mainly just asking for doc improvements.

ambron60:
Greetings.. what is this 23.217.138.110 Akamai host and what is p 22 being used for?

v4/core/base_service.go
padamstx: Yes, at least for now, disable-ssl and retries are mutually exclusive.   I think there might be a way to support both simultaneously.   I can explore that as a follow-on.   For now, I'll document the current situation.

v4/core/base_service.go
```diff
@@ -451,3 +493,157 @@ func getErrorMessage(responseMap map[string]interface{}, statusCode int) string
...
+type httpLogger struct {
+}
+
+func (l *httpLogger) Printf(format string, inserts ...interface{}) {
+	GetLogger().Log(LevelDebug, format, inserts...)
+}
+
+// NewRetryableHTTPClient returns a new instance of go-retryablehttp.Client
+// with a default configuration that supports Go SDK usage.
+func NewRetryableHTTPClient() *retryablehttp.Client {
```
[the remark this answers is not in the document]
padamstx: I initially made this private, thinking that EnableRetries() would be the only way that an SDK user would be able to enable retries.   But then I decided to make this public to give the SDK user more freedom in configuring a retryable client for rare cases where EnableRetries() isn't enough to satisfy their requirements.  For example, if the SDK user wanted to use their own "CheckRetry" and/or "Backoff" callback functions, they can use NewRetryableHTTPClient(), then set the various fields however they want.   With the way in which the retryable client is hung off the BaseService.Client field (i.e. it's hidden inside the BaseService.Client's Transport field), it would be difficult for an SDK user to get ahold of that after calling EnableRetries().

padamstx:
> Greetings.. what is this 23.217.138.110 Akamai host and what is p 22 being used for?

Hi @ambron60, not sure what this question is pertaining to. Could you be more specific?

ambron60:
Hi @padamstx and thanks for taking my question.. one of our hosts was resolving to this particular IP and I can't find any info/reference to it except for its mapping to Akamai- I saw you guys post above making reference to it, so I was inclined to ask.. it also looks like it has ports 22 and 80 open, not sure why.

Ex IBM-er here btw.. 👍

padamstx:
> Hi @padamstx and thanks for taking my question.. one of our hosts was resolving to this particular IP and I can't find any info/reference to it except for its mapping to Akamai- I saw you guys post above making reference to it, so I was inclined to ask.. it also looks like it has ports 22 and 80 open, not sure why.

Ah, now I see the connection.   I think it was just a coincidence that one of our testcases uses hostname "notahost" in a negative test, and that apparently resolved to the IP address, but I'm not sure how that happened to be honest.   This PR really has no connection to Akamai at all.

ambron60:
> > Hi @padamstx and thanks for taking my question.. one of our hosts was resolving to this particular IP and I can't find any info/reference to it except for its mapping to Akamai- I saw you guys post above making reference to it, so I was inclined to ask.. it also looks like it has ports 22 and 80 open, not sure why.
> 
> Ah, now I see the connection. I think it was just a coincidence that one of our testcases uses hostname "notahost" in a negative test, and that apparently resolved to the IP address, but I'm not sure how that happened to be honest. This PR really has no connection to Akamai at all.

Thanks!!

PR #83, opened 2020-10-28
chore: appscan annotations

christiancompton:
This PR adds exceptions using the [gosec annotations](https://github.com/securego/gosec#annotating-code) to ignore false positives in security scans.


padamstx (approved):
LGTM

PR #84, opened 2020-10-29
fix: support enable-retries via external configuration

padamstx:
This PR adds support for enabling retries via external config properties.
I completely whiffed on supporting external config properties for the new retries feature in my original PR, and discovered that while working on the documentation yesterday.


jorge-ibm (approved):
LGTM

PR #86, opened 2020-11-16
fix: improve serialization of DateTime values

padamstx:
Fixes arf/planning-sdk-squad#2313

This commit includes new function and tests to ensure
that the Go core can support all of the (currently
understood) required date-time parsing formats for
IBM Cloud services.
It also includes a change to ensure that a DateTime
value's UTC representation is used "on the wire".


mkistler (approved):
Looks good! 👍

PR #87, opened 2020-12-03
feat(BaseService): add Clone() method to clone a BaseService instance

padamstx:
This commit adds the BaseService "Clone()" method that can be
used to make a copy of an existing BaseService instance as part
of cloning a generated service struct instance.


mkistler (approved):
Looks good! 👍

PR #90, opened 2021-01-17
Minor improvement to logging implementation

mkistler:
This PR makes a small improvement to the `SetLoggingLevel` method to only completely replace the current logger in the (unlikely) situation where the logger is a "custom" implementation of the `Logger` class, and not an `SDKLoggerImpl`. The prior behavior of replacing the logger just to set the logging level is not what most users would expect from a "Set" method.

In the next major release, we should add `SetLogLevel` to the `Logger` interface to make the behavior of `SetLoggingLevel` consistent for custom `Logger` implementations.

### DIFF
--- a/v4/core/log.go
+++ b/v4/core/log.go
@@ -36,6 +36,8 @@ const (
 // Users of the library can supply their own implementation by calling SetLogger().
 type Logger interface {
 	Log(level LogLevel, format string, inserts ...interface{})
+	// TODO: add in next major release
+	//SetLogLevel(level LogLevel)
 	Error(format string, inserts ...interface{})
 	Warn(format string, inserts ...interface{})
 	Info(format string, inserts ...interface{})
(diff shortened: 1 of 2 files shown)


v4/core/log.go
```diff
@@ -36,6 +36,8 @@ const (
 // Users of the library can supply their own implementation by calling SetLogger().
 type Logger interface {
 	Log(level LogLevel, format string, inserts ...interface{})
+	// TODO: add in next major release
```
padamstx, line 39: In @dpopp07 's go core PR  re: logging, I recommended some other methods be added to the Logger interface as well, but completely whiffed on the fact that it would introduce a breaking change.
I think we'll need to add methods to the Logger interface in order to solve Dustin's issue the "right way".
We could avoid creating a major version and take a chance that nobody has implemented their own Logger impl (probably not preferable), but it wouldn't be all that strenuous to just create a new major version (v5) of the go core (and introduce SetLogLevel() and any changes from Dustin's PR, then update the generator (as part of Dustin's logging-related generator PR) to generate Go code that references go core v5 instead of v4.

padamstx (approved):
LGTM

padamstx:
Closing this PR as it was superceded by https://github.com/IBM/go-sdk-core/pull/91

PR #91, opened 2021-01-19
feat: add debug logging of requests/responses

padamstx:
This commit adds debug logging of outbound requests and inbound
responses.  These will be logged if log level "debug" is enabled.

BREAKING CHANGE: several methods added to Logger interface

Several methods were added to the Go core's Logger interface:
SetLogLevel(), GetLogLevel(), and IsLogLevelEnabled().
These additional methods will need to be added to any
user implementations of the Logger interface.

BREAKING CHANGE: additional parameter added to NewLogger() signature

The NewLogger() function has a new parameter "errorLogger".
Any calls to NewLogger() will need to be modified to include the
new parameter.

BREAKING CHANGE: deprecated unmarshal-related methods have been removed

Several deprecated unmarshal-related methods were removed from the Go core:
- UnmarshalString, UnmarshalStringSlice, UnmarshalStringMap, UnmarshalStringMapSlice
- UnmarshalByteArray, UnmarshalByteArraySlice, UnmarshalByteArrayMap, UnmarshalByteArrayMapSlice
- UnmarshalBool, UnmarshalBoolSlice, UnmarshalBoolMap, UnmarshalBoolMapSlice
- UnmarshalInt64, UnmarshalInt64Slice, UnmarshalInt64Map, UnmarshalInt64MapSlice
- UnmarshalFloat32, UnmarshalFloat32Slice, UnmarshalFloat32Map, UnmarshalFloat32MapSlice
- UnmarshalFloat64, UnmarshalFloat64Slice, UnmarshalFloat64Map, UnmarshalFloat64MapSlice
- UnmarshalUUID, UnmarshalUUIDSlice, UnmarshalUUIDMap, UnmarshalUUIDMapSlice
- UnmarshalDate, UnmarshalDateSlice, UnmarshalDateMap, UnmarshalDateMapSlice
- UnmarshalDateTime, UnmarshalDateTimeSlice, UnmarshalDateTimeMap, UnmarshalDateTimeMapSlice
- UnmarshalObject, UnmarshalObjectSlice
- UnmarshalAny, UnmarshalAnySlice, UnmarshalAnyMap, UnmarshalAnyMapSlice
These methods are no longer used by code emitted by the Go generator.  If you
have old generated Go code that still uses these methods, then you should continue
using version 4 of the Go core, or regenerate your SDK code using a new version of the
SDK generator.


padamstx:
I think that is the longest commit message I've ever written :1st_place_medal:

v5/core/log.go
```diff
@@ -0,0 +1,202 @@
...
+	Debug(format string, inserts ...interface{})
+
+	SetLogLevel(level LogLevel)
+	GetLogLevel() LogLevel
+
+	IsLogLevelEnabled(level LogLevel) bool
+	IsErrorEnabled() bool
+	IsWarnEnabled() bool
+	IsInfoEnabled() bool
+	IsDebugEnabled() bool
```
mkistler, line 51: I wonder ... is there some way we can implement these methods ourselves? I mean ... as long as we have `GetLogLevel`, then the implementations of these other methods are straightforward, so why require the user to write these if they provide a custom logger?

In fact, I would go further.  We _should not allow_ the user to implement these methods, just to avoid the potential for them screwing it up.  Can we do that?

v5/core/log.go
```diff
@@ -0,0 +1,202 @@
...
+}
+
+// SDKLoggerImpl is the Go core's implementation of the Logger interface.
+// This logger contains two instances of Go's log.Logger interface which are
+// used to perform message logging.  One instance represents "stdout" and other
+// represents "stderr".  Messages logged at level Error or Debug will appear on
+// the "stderr" logger, while all other messages will appear on the "stdout" logger.
```
mkistler, line 58: I suppose this is fine.  In the event that a user wants to see Error and Debug messages interleaved with Warning and Info messages, then that user just needs to pipe both stdout and stderr to the same output.

But it does seem a bit arbitrary how we've chosen to map severities to these two streams.  I mean, I could easily imagine that some users might want:
- Error and Warn -> stderr
- Info and Debug -> stdout

and other mappings are possible.

Again -- this is fine.  But if there was a way to let the user choose how to map severities to streams, without over-complicating things, it might be worth considering.

mkistler (approved):
Looks good. 👍 

I left a few "musings" ... things to consider ... but I leave it to your judgement on whether any changes are needed.

v5/core/log.go
padamstx: If we were doing this in Java, we'd implement common behavior in a base class and then each implementation could subclass that.  But in Go, there's no true inheritance.  I suppose one could use composition and "inherit" some type we define with the common behaviors, but to be honest I'm not sure anyone is even implementing their own Logger's.

When making these updates I perhaps leaned a bit toward exposing more methods rather than less, to avoid another situation like we're fixing now :)
I could remove the Is[Error,Warn,Info,Debug]Enabled() methods and just leave IsLogLevelEnabled(LogLevel) there.

As for your last question, I don't know off-hand how we'd prevent someone from implementing specific methods, but if someone is implementing their own Logger they'd need to implement all the methods of the interface, so...  maybe I didn't understand the question correctly :)

v5/core/log.go
padamstx: I used this mapping of loglevel -> stream to accommodate the CLI plugin situation where we don't want anything written to stdout under normal situations because that's where the response objects are written.
However, typing that previous sentence caused me to re-think that strategy so I'll need to discuss with @dpopp07 again :)

v5/core/log.go
padamstx: Addressed this by removing the Is[Error,Warn,Info,Debug]Enabled() methods in favor of just IsLogLevelEnabled().

v5/core/log.go
padamstx: I discussed this with Dustin and I changed the SDKLoggerImpl type slightly to focus on the types of messages being logged rather than the particular stream.   So now we have "infoLogger" and "errorLogger".  Those will map to stdout and stderr respectively, but can easily be changed when calling NewLogger().  For example, Dustin will initialize a Logger instance that uses stderr for both sets of messages in order to avoid having logged messages interfere with operation responses inthe CLI environment.

dpopp07: @padamstx I think this comment needs to change with the updated implementation, right?

dpopp07 (approved):
Looks good! Just left one note where I think you need to update your comment to reflect your updated implementation

padamstx:
> I think you need to update your comment to reflect your updated implementation

Good catch.  I updated the comment.   I fixed it in one place initially, but missed the second place :)

PR #95, opened 2021-02-18
fix: ensure result value is set if err is nil

mkistler:
This PR addresses a corner case in the base_service `Request` method where it could return a nil `err` but not set the `result` value passed by the user.  This behavior would force users to check explicitly for `result != nil` in addition to checking for `err == nil`, which is a poor experience for the user, particularly since they will likely just manufacture an error for this case anyway.

I also made a small change to check for a nil input to `UnmarshallModel`, since this method should treat this condition as an error rather than silently returning and empty result as it had been.

Tests are updated.

### DIFF
--- a/v5/core/base_service_test.go
+++ b/v5/core/base_service_test.go
@@ -252,7 +252,7 @@ func TestRequestGoodResponseText(t *testing.T) {
 	assert.Nil(t, err)
 	assert.NotNil(t, service.Options.Authenticator)
 	assert.Equal(t, AUTHTYPE_NOAUTH, service.Options.Authenticator.AuthenticationType())
-	detailedResponse, err := service.Request(req, new(string))
+	detailedResponse, err := service.Request(req, new([]byte))
 	assert.Nil(t, err)
 	assert.NotNil(t, detailedResponse)
 	assert.Equal(t, "text/plain", detailedResponse.GetHeaders().Get("Content-Type"))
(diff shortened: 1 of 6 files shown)


padamstx:
@mkistler Looks like you have some unit test failures due to your changes (see travis builds) and your feature branch is out of date also.   Make sure you're basing your work on the new "main" branch :)

v5/core/base_service_test.go
```diff
@@ -252,7 +252,7 @@ func TestRequestGoodResponseText(t *testing.T) {
 	assert.Nil(t, err)
 	assert.NotNil(t, service.Options.Authenticator)
 	assert.Equal(t, AUTHTYPE_NOAUTH, service.Options.Authenticator.AuthenticationType())
-	detailedResponse, err := service.Request(req, new(string))
+	detailedResponse, err := service.Request(req, new([]byte))
```
padamstx, line 255: I think for a text response, we'd expect the operation's return value to be `*string` (a string ptr).
```suggestion
        var result *string
	detailedResponse, err := service.Request(req, &result)
```
then we should probably verify with an assert that we got back the expected response string.
Looks like the test might have been at least slightly incorrect to begin with.

v5/core/base_service_test.go
```diff
@@ -705,7 +773,7 @@ func TestRequestBasicAuth1(t *testing.T) {
 	builder.AddQuery("Version", "2018-22-09")
 	req, _ := builder.Build()
 
-	var foo *Foo
+	var foo []byte
 	_, err = service.Request(req, &foo)
```
padamstx, line 777:
```suggestion
	_, err = service.Request(req, nil)
```

v5/core/base_service_test.go
```diff
@@ -747,7 +815,7 @@ func TestRequestBasicAuth2(t *testing.T) {
 	assert.NotNil(t, service)
 	assert.NotNil(t, service.Options.Authenticator)
 
-	var foo *Foo
+	var foo []byte
 	_, err = service.Request(req, &foo)
```
padamstx, line 819:
```suggestion
	_, err = service.Request(req, nil)
```
For this test, it looks like we're not actually sending a response body back, so should probably just pass in nil for the result param to the Request() method, no?

v5/core/base_service_test.go
```diff
@@ -798,7 +866,7 @@ func TestRequestNoAuth1(t *testing.T) {
 	assert.NotNil(t, service.Options.Authenticator)
 	assert.Equal(t, AUTHTYPE_NOAUTH, service.Options.Authenticator.AuthenticationType())
 
-	var foo *Foo
+	var foo []byte
 	_, err = service.Request(req, &foo)
```
padamstx, line 870:
```suggestion
	_, err = service.Request(req, nil)
```

v5/core/base_service_test.go
```diff
@@ -834,7 +902,7 @@ func TestRequestNoAuth2(t *testing.T) {
 	assert.Nil(t, err)
 	assert.Equal(t, AUTHTYPE_NOAUTH, service.Options.Authenticator.AuthenticationType())
 
-	var foo *Foo
+	var foo []byte
 	_, err = service.Request(req, &foo)
```
padamstx, line 906:
```suggestion
	_, err = service.Request(req, nil)
```

v5/core/base_service_test.go
```diff
@@ -881,7 +949,7 @@ func TestRequestIAMAuth(t *testing.T) {
 	assert.NotNil(t, service.Options.Authenticator)
 	assert.Equal(t, AUTHTYPE_IAM, service.Options.Authenticator.AuthenticationType())
 
-	var foo *Foo
+	var foo []byte
 	_, err = service.Request(req, &foo)
```
padamstx, line 953:
```suggestion
	_, err = service.Request(req, nil)
```

v5/core/base_service_test.go
```diff
@@ -1013,7 +1081,7 @@ func TestRequestIAMWithIdSecret(t *testing.T) {
 	assert.NotNil(t, service)
 	assert.NotNil(t, service.Options.Authenticator)
 
-	var foo *Foo
+	var foo []byte
 	_, err = service.Request(req, &foo)
```
padamstx, line 1085:
```suggestion
	_, err = service.Request(req, nil)
```

v5/core/unmarshal_v2_models_test.go
```diff
@@ -173,6 +173,11 @@ func TestUnmarshalModelInstance(t *testing.T) {
 	assert.NotNil(t, myModel)
 	assert.Equal(t, "string1", *(myModel.Foo))
 	assert.Equal(t, int64(44), *(myModel.Bar))
+
+	// Unmarshal with "zero" input should return an error
+	var rawMap2 map[string]json.RawMessage
+	err = UnmarshalModel(rawMap2, "", &myModel, UnmarshalMyModel)
+	assert.NotNil(t, err)
```
padamstx, line 180: we should probably also have a test that exercises an un-initialized []RawMessage as well

padamstx (approved):
I suggested some changes to some of the tests which do not actually have a response body.  I think the previous version of those tests were incorrect (they should have been supplying nil for the result param passed to BaseService.Request()), but that test "bug" was not exposed due to the prior behavior of the Request() function.
I'll approve now to avoid another review cycle.

v5/core/base_service_test.go
mkistler: If we make this change, then this test becomes redundant with the following test, `TestRequestGoodResponseString` -- both return a string response with `text/plain` content-type, and only differ in the type of `result`.  I accepted your suggestion, but now I think I will revert back to my original code unless you think I've missed something here.

v5/core/unmarshal_v2_models_test.go
mkistler: Added.

PR #96, opened 2021-02-22
feat: add UUID, date, and datetime helpers for terraform usage

rmkeezer:
Add helper methods for casting strfmt UUID, Date, and DateTime literals to pointers needed by the go setter methods. Also, this is only parsing dates with RFC3339. This likely needs to be updated in the future to allow other formats to be used.

This is used in the sdk generator PR #873

### DIFF
--- a/v5/core/utils.go
+++ b/v5/core/utils.go
@@ -26,8 +26,8 @@ import (
 	"strings"
 	"time"
 
-	validator "gopkg.in/go-playground/validator.v9"
 	"github.com/go-openapi/strfmt"
+	validator "gopkg.in/go-playground/validator.v9"
 )
 
 // Validate is a shared validator instance used to perform validation of structs.
@@ -186,27 +186,6 @@ func PrettyPrint(result interface{}, resultName string) {
 // GetCurrentTime returns the current Unix time.
 func GetCurrentTime() int64 {
 	return time.Now().Unix()
-}
-
-// ParseDate parses the specified RFC3339 full-date string (YYYY-MM-DD) and returns a strfmt.Date instance.
-func ParseDate(dateString string) (fmtDate strfmt.Date, err error) {
-	formattedTime, err := time.Parse(strfmt.RFC3339FullDate, dateString)
-	var formattedDate strfmt.Date
-	if err != nil {
-		return formattedDate, err
-	}
-	formattedDate = strfmt.Date(formattedTime)
-	return formattedDate, nil
-}
-
-// ParseDateTime parses the specified date-time string and returns a strfmt.DateTime instance.
-func ParseDateTime(dateString string) (fmtDTime strfmt.DateTime, err error) {
-	var formattedDateTime strfmt.DateTime
-	formattedDateTime, err = strfmt.ParseDateTime(dateString)
-	if err != nil {
-		return formattedDateTime, err
-	}
-	return formattedDateTime, nil
 }
 
 // ConvertSlice Marshals 'slice' to a json string, performs
--- a/v5/core/utils.go
+++ b/v5/core/utils.go
@@ -27,6 +27,7 @@ import (
 	"time"
 
 	validator "gopkg.in/go-playground/validator.v9"
+	"github.com/go-openapi/strfmt"
 )
 
 // Validate is a shared validator instance used to perform validation of structs.
@@ -109,6 +110,11 @@ func Float64Ptr(literal float64) *float64 {
 	return &literal
 }
 
+// UUIDPtr returns a pointer to strfmt.UUID literal.
+func UUIDPtr(literal strfmt.UUID) *strfmt.UUID {
+	return &literal
+}
+
 // IsJSONMimeType Returns true iff the specified mimeType value represents a
 // "JSON" mimetype.
 func IsJSONMimeType(mimeType string) bool {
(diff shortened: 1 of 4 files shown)


v5/core/utils.go
```diff
@@ -182,6 +188,20 @@ func GetCurrentTime() int64 {
...
+// ParseDate parses and returns a strfmt Date pointer
+func ParseDate(dateString string) *strfmt.Date {
+	formattedTime, _ := time.Parse("2006-01-02", dateString)
+	formattedDate := strfmt.Date(formattedTime)
+	return &formattedDate
+}
+
+// ParseDateTime parses and returns a strfmt DateTime pointer
+func ParseDateTime(dateString string) *strfmt.DateTime {
+	formattedTime, _ := time.Parse(time.RFC3339, dateString)
```
mkistler, line 200: You should use `strfmt.ParseDateTime` here.  It has already been set up to handle all the date-time formats we need.

v5/core/utils.go
```diff
@@ -182,6 +188,20 @@ func GetCurrentTime() int64 {
 	return time.Now().Unix()
 }
 
+// ParseDate parses and returns a strfmt Date pointer
+func ParseDate(dateString string) *strfmt.Date {
```
mkistler, line 192: It seems like we should have an error return on this method.

v5/core/utils.go
```diff
@@ -182,6 +188,20 @@ func GetCurrentTime() int64 {
...
+	formattedTime, _ := time.Parse("2006-01-02", dateString)
+	formattedDate := strfmt.Date(formattedTime)
+	return &formattedDate
+}
+
+// ParseDateTime parses and returns a strfmt DateTime pointer
+func ParseDateTime(dateString string) *strfmt.DateTime {
```
mkistler, line 199: I think we should have an error return on this method.

mkistler (changes requested):
I think the date/time conversion methods should have error returns and use the date-time conversion support we have already built into the core.

v5/core/utils.go
rmkeezer: Updated to use `strfmt.ParseDateTime` with error handling!

v5/core/utils.go
rmkeezer: Updated with an error return!

v5/core/utils_test.go
```diff
@@ -318,6 +321,19 @@ func TestConvertSliceByteArray(t *testing.T) {
 	assert.Empty(t, convertedSlice)
 }
 
+func TestDateTimeUtil(t *testing.T) {
+	dateVar := strfmt.Date(time.Now())
+	fmtDate, err := ParseDate(dateVar.String())
+	assert.Nil(t, err)
+	assert.True(t, dateVar.String() == fmtDate.String())
```
padamstx, line 328:
```suggestion
	assert.Equal(t, dateVar.String(), fmtDate.String())
```

v5/core/utils_test.go
```diff
@@ -318,6 +321,19 @@ func TestConvertSliceByteArray(t *testing.T) {
...
+
+	dateTimeVar := strfmt.DateTime(time.Now())
+	var fmtDTime *strfmt.DateTime
+	fmtDTime, err = ParseDateTime(dateTimeVar.String())
+	assert.Nil(t, err)
+	assert.True(t, dateTimeVar.String() == fmtDTime.String())
```
padamstx, line 334:
```suggestion
	assert.Equal(t, dateTimeVar.String(), fmtDTime.String())
```

v5/core/utils.go
```diff
@@ -182,6 +188,25 @@ func GetCurrentTime() int64 {
 	return time.Now().Unix()
 }
 
+// ParseDate parses and returns a strfmt Date pointer
```
padamstx, line 191:
```suggestion
// ParseDate parses the specified RFC3339 full-date string (YYYY-MM-DD) and returns a strfmt.Date instance.
```

v5/core/utils.go
```diff
@@ -182,6 +188,25 @@ func GetCurrentTime() int64 {
...
+func ParseDate(dateString string) (fmtDate *strfmt.Date, err error) {
+	formattedTime, err := time.Parse(strfmt.RFC3339FullDate, dateString)
+	if err != nil {
+		return nil, err
+	}
+	formattedDate := strfmt.Date(formattedTime)
+	return &formattedDate, nil
+}
+
+// ParseDateTime parses and returns a strfmt DateTime pointer
```
padamstx, line 201:
```suggestion
// ParseDateTime parses the specified date-time string and returns a strfmt.DateTime instance.
```

padamstx (changes requested):
Looks good so far.  Just had some minor comments.  Also, I think it would be good to add at least one negative test for ParseDate() and ParseDateTime().

padamstx (approved):
LGTM

v5/core/utils.go
```diff
@@ -182,6 +188,25 @@ func GetCurrentTime() int64 {
...
+	formattedTime, err := time.Parse(strfmt.RFC3339FullDate, dateString)
+	if err != nil {
+		return nil, err
+	}
+	formattedDate := strfmt.Date(formattedTime)
+	return &formattedDate, nil
+}
+
+// ParseDateTime parses the specified date-time string and returns a strfmt.DateTime instance.
+func ParseDateTime(dateString string) (fmtDTime *strfmt.DateTime, err error) {
```
mkistler, line 202: I think it would be better for this method to return a `strfmt.DateTime` rather than a pointer to `strfmt.DateTime`.  This is perhaps a question of style, and thus somewhat subjective, but in my defense I will note that `strfmt.ParseDateTime` returns a `strfmt.DateTime`.

v5/core/utils.go
```diff
@@ -182,6 +188,25 @@ func GetCurrentTime() int64 {
 	return time.Now().Unix()
 }
 
+// ParseDate parses the specified RFC3339 full-date string (YYYY-MM-DD) and returns a strfmt.Date instance.
+func ParseDate(dateString string) (fmtDate *strfmt.Date, err error) {
```
mkistler, line 192: Here also, I think we should return a `strfmt.Date` rather than a pointer.

mkistler (changes requested):
This looks good but I'd like a small change in the interface.

mkistler (approved):
Looks good! 👍

PR #97, opened 2021-02-23
fix(IAM Authenticator): canonicalize iam url & improve iam error reporting

padamstx:
This commit includes two changes:
1. The IAM Authenticator is now more flexible with respect to the
configured IAM url specified by the user.  The path part ("/identity/token")
is now optional.
2. For errors returned by the IAM token server, if the response received
from the token server does not contain an error message, then a default
error message containing the unexpected status code will be returned rather
than an empty string.


mkistler (approved):
Looks good! 👍

PR #98, opened 2021-02-23
Add GetQueryParam utility method to support pagination

mkistler:
This PR adds a small utility method to extract the value of a query parameter from a URL string.  This will be used in the pagination support for Go.

### DIFF
--- a/v5/core/utils.go
+++ b/v5/core/utils.go
@@ -18,6 +18,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"net/url"
 	"os"
 	"reflect"
 	"regexp"
@@ -260,4 +261,28 @@ func SliceContains(slice []string, contains string) bool {
 		}
 	}
 	return false
+}
+
+// return a pointer to the value of query parameter `param` from url,
+// or nil when not found
+func GetQueryParam(urlStr *string, param string) (*string, error) {
+	if urlStr == nil || *urlStr == "" {
+		return nil, nil
+	}
+
+	u, err := url.Parse(*urlStr)
+	if err != nil {
+		return nil, err
+	}
+
+	q, err := url.ParseQuery(u.RawQuery)
+	if err != nil {
+		return nil, err
+	}
+
+	v := q.Get(param)
+	if v == "" {
+		return nil, nil
+	}
+	return &v, nil
 }
--- a/v5/core/utils_test.go
+++ b/v5/core/utils_test.go
@@ -480,3 +480,55 @@ func TestSliceContains(t *testing.T) {
 
 	assert.False(t, SliceContains(nil, "foo"))
 }
+
+func TestGetQueryParam(t *testing.T) {
+	nextURL := "/api/v1/offerings?start=foo&limit=10"
+	next, err := GetQueryParam(&nextURL, "start")
+	assert.Nil(t, err)
+	assert.Equal(t, "foo", *next)
+
+	fqNextURL := "https://acme.com/api/v1/offerings?start=bar&limit=10"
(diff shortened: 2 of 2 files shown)


mkistler:
I've marked this as a `draft` because I think we should hold off merging until the rest of the pagination support is worked out, but I wanted to post this to get feedback on the design/implementation.

v5/core/utils.go
```diff
@@ -255,3 +256,18 @@ func SliceContains(slice []string, contains string) bool {
 	}
 	return false
 }
+
```
padamstx, line 259: For reference, the platform services Go SDK has a few utility methods scattered among the integration tests:
* https://github.com/IBM/platform-services-go-sdk/blob/74f8fd8344851feceeb0a5ebc09b10d0a16c67cb/usermanagementv1/user_management_v1_integration_test.go#L306
* https://github.com/IBM/platform-services-go-sdk/blob/74f8fd8344851feceeb0a5ebc09b10d0a16c67cb/iamidentityv1/iam_identity_v1_integration_test.go#L583
* https://github.com/IBM/platform-services-go-sdk/blob/74f8fd8344851feceeb0a5ebc09b10d0a16c67cb/enterpriseusagereportsv1/enterprise_usage_reports_v1_integration_test.go#L240
They all pretty much work the same way.  We can replace those with the new core function when it's available

v5/core/utils.go
```diff
@@ -255,3 +256,18 @@ func SliceContains(slice []string, contains string) bool {
...
+	if IsNil(urlStr) {
+		return "", nil
+	}
+
+	u, err := url.Parse(*urlStr)
+	if err != nil {
+		return "", err
+	}
+
+	q := u.Query()
```
padamstx, line 271: One main difference that you'll notice in the methods I linked to above is the use of the url's `RawQuery` field and calling the `ParseQuery()` method.  I don't remember the significance right now, but there must have been a reason I coded it like that - LOL.

padamstx (approved):
The changes look fine so far, although the differences between your new function and the ones used in the Go SDK might be important, but not 100% sure.
I think when all this pagination stuff gels a little more, we have 4 SDK projects that are just itching to be first in line to try it out :)

v5/core/utils.go
mkistler: I think this could make a difference if a) there is no query string, or b) if the query string exists but is in some way invalid.  I'll try to write some tests for these cases to see how my implementation handles them.

mkistler: Thanks for the feedback Phil. I've modified `GetQueryParam` to be more like the Platform Services methods, though a bit more concise (style preference) and updated the tests to cover all the cases (I think).

mkistler:
One more variation to add to the tests: repeated query param.

v5/core/utils_test.go
```diff
@@ -480,3 +480,55 @@ func TestSliceContains(t *testing.T) {
...
+
+	fqNextURL := "https://acme.com/api/v1/offerings?start=bar&limit=10"
+	next, err = GetQueryParam(&fqNextURL, "start")
+	assert.Nil(t, err)
+	assert.Equal(t, "bar", *next)
+
+	// No query parameter
+	next, err = GetQueryParam(&nextURL, "token")
+	assert.Nil(t, err)
+	assert.Equal(t, (*string)(nil), next)
```
padamstx, line 498:
```suggestion
	assert.Nil(t, next)
```
Unless there's a good reason for using assert.Equal() in these cases, I think you could just use assert.Nil() instead here and in the other nil checks below.

padamstx (approved):
Looks good.  I just had a suggestion to use assert.Nil() in a few places in the tests.

PR #102, opened 2021-03-25
fix: data race warnings

pyrooka:
Fixes arf/planning-sdk-squad#2440

## Problem
There were several warnings about race conditions (mostly about the `tokenData` field) when running the unit tests. 

## Reproduce
`go test -tags=all -race`

## Solution
Introduced "getter" and "setter" functions for the authenticator's `tokenData` with mutex.

### DIFF
--- a/v5/core/base_service_test.go
+++ b/v5/core/base_service_test.go
@@ -1497,9 +1497,7 @@ func TestAuthNotConfigured(t *testing.T) {
 }
 
 func TestAuthInterfaceNilValue(t *testing.T) {
-	// Verify a non-nil service is cloned correctly.
-	authenticator := &NoAuthAuthenticator{}
-	authenticator = nil
+	var authenticator *NoAuthAuthenticator = nil
 
 	options := &ServiceOptions{
 		Authenticator: authenticator,
--- a/v5/core/base_service_test.go
+++ b/v5/core/base_service_test.go
@@ -1496,6 +1496,21 @@ func TestAuthNotConfigured(t *testing.T) {
 	assert.Nil(t, service)
 }
 
+func TestAuthInterfaceNilValue(t *testing.T) {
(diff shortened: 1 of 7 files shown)


v5/core/cp4d_authenticator.go
```diff
@@ -129,15 +129,15 @@ func newCloudPakForDataAuthenticatorFromMap(properties map[string]string) (*Clou
 }
 
 // AuthenticationType returns the authentication type for this authenticator.
-func (CloudPakForDataAuthenticator) AuthenticationType() string {
```
pyrooka, line 129: I don't know if is there a specific reason why this used as a value or not, but hopefully changing it to pointer won't break anything. All tests passed btw.

@padamstx  What's your take on this?

padamstx (approved):
LGTM

v5/core/base_service_test.go
```diff
@@ -1496,6 +1496,21 @@ func TestAuthNotConfigured(t *testing.T) {
 	assert.Nil(t, service)
 }
 
+func TestAuthInterfaceNilValue(t *testing.T) {
+	// Verify a non-nil service is cloned correctly.
+	authenticator := &NoAuthAuthenticator{}
+	authenticator = nil
```
padamstx, line 1502: this actually causes a lint error in the travis build :)
you could just pass nil directly in the struct literal below, or change the above 2 lines to be something like this:
```
        var authenticator *NoAuthAuthenticator = nil
```

padamstx (approved):
After approving the PR, I checked on the travis build and noticed the lint error, so added a new review with comment.   I'll approve now to avoid a re-review.

v5/core/base_service_test.go
pyrooka: Sorry, I forgot to run the linter in local! Fixed now.

PR #103, opened 2021-03-28
feat: add support for unmarshalling two-dimensional slices of model instances

padamstx:
This commit improves the model unmarshalling capabilities within the Go core
by adding support for an additional pattern ([][]MyModel) within the
UnmarshalModel() function.  The full set of patterns now supported are:
- MyModel (model instance)
- map[string]MyModel (map containing model instances)
- map[string][]MyModel (map contains slices of model instances
- []MyModel (slice of model instances)
- new: [][]MyModel (slice of model instance slices)

### DIFF
--- a/v5/core/unmarshal_v2_primitives_test.go
+++ b/v5/core/unmarshal_v2_primitives_test.go
@@ -33,6 +33,7 @@ func TestUnmarshalPrimitiveString(t *testing.T) {
 	type MyModel struct {
 		Prop              *string
 		PropSlice         []string
+		PropSliceSlice    [][]string
 		PropMap           map[string]string
 		PropSliceMap      map[string][]string
 		PropMapSlice      []map[string]string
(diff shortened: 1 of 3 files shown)


padamstx:
@mkistler @pyrooka I still need to perform some testing with the Go generator and with the platform-services SDK project, but I think this PR is in shape to begin reviewing.

v5/core/unmarshal_v2_primitives_test.go
```diff
@@ -50,13 +50,15 @@ func TestUnmarshalPrimitiveString(t *testing.T) {
 		"bad_type":  true,
```
mkistler, line 50: Should there be a variation in this test for `[][]string`?  That is supposed to work, right?

padamstx:
@mkistler I've pushed a new commit that adds tests for two-dimensional arrays of primitive types.

mkistler (approved):
Looks good! 👍

PR #109, opened 2021-04-21
fix: eliminate goroutine leak in the authenticators

### LINKED ISSUE #108: goroutine leak core.IamAuthenticator
There appears to be a goroutine leak happening in [iam_authenticator.go](https://github.com/IBM/go-sdk-core/blob/main/v5/core/iam_authenticator.go#L216).

The function is spawning a goroutine which will then idle on the sending to the channel if the main goroutine executes the select statement before `authenticator.invokeRequestTokenData()` returns. The problem is `default` on L225 will exit the select statements scope before receiving the new token.

Profile:
[leak.pdf](https://github.com/IBM/go-sdk-core/files/6344609/leak.pdf)

pyrooka:
This PR fixes some possible goroutine leak in the authenticators' `getToken` methods.

Closes #108


padamstx (approved):
LGTM

PR #110, opened 2021-04-21
feat: add FileWithMetadata type to the core

dpopp07:
FileWithMetadata is a struct that contains the binary data within a file, along
with specific metadata: the filename and the content type. This is a type that
the generated code occasionally uses. Currently, we generate this type as a
model on a specific service but because it is service-independent, we want to
move it to the core. This adds it as a supported type within the core. Although
the generator won't be able to change this type in options models until the next
major relase, it will use this type in the generated unmarshall functions, which
are currently broken. This will be fixed in the generator after this feature is
released.

The new "unmarshaler" in the core includes code to parse the JSON
string and read the data from the file while creating the struct. Generated
unmarshaller on service structs will invoke this new, working unmarshaller.

I have verified that this works and this PR will be a necessary dependency of my upcoming generator PR.

### DIFF
--- a/v5/core/file_with_metadata_test.go
+++ b/v5/core/file_with_metadata_test.go
@@ -1,0 +1,92 @@
+package core
+
+// (C) Copyright IBM Corp. 2021.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import (
+	"bytes"
+	"encoding/json"
+	"github.com/stretchr/testify/assert"
+	"io"
+	"io/ioutil"
+	"os"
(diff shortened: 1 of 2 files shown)


padamstx:
@dpopp07 Just noticed the builds for this PR failed due to the use of the testing.T.TempDir() function.   That was apparently introduced in Go 1.15 and the minimum version of Go that we currently support in the core and SDKs is 1.12.   Is there an alternative that you can use instead???

v5/core/file_with_metadata_test.go
```diff
@@ -0,0 +1,85 @@
...
+
+	assert.Nil(t, model.Filename)
+	assert.Nil(t, model.ContentType)
+}
+
+func TestUnmarshalFileWithMetadata(t *testing.T) {
+	var err error
+
+	// setup the test by creating a temp file for the unmarshaler to read
+	tempdir := t.TempDir() // this is automatically cleaned up
```
padamstx, line 60: Need to code around the fact that TempDir() is not present until Go 1.15, but we currently support Go 1.12+

padamstx (changes requested):
Looks good except for the TempDir() issue :)

v5/core/file_with_metadata_test.go
dpopp07: Ah rats, I didn't notice that. I didn't realize I had Go 1.15 locally either but I guess I do. I can go back to what I was doing before I found that function and create my own temp dir

padamstx (approved):
LGTM

PR #111, opened 2021-04-23
fix: handle missing response body

padamstx:
This PR contains changes to the BaseService.Request() function to handle scenarios where we expect a response body for an operation invocation but one was not returned by the server.   This might occur in situations where an operation could return multiple success status codes, some with and some without a response body (think 202 with a response body and a 204 without a response body).
This new behavior aligns with the current behavior of the Java core where a null result object would be returned to the SDK user if a response was expected but no response body was received.


mkistler (approved):
Looks good! 👍

PR #114, opened 2021-04-29
fix: update JWT package

### LINKED ISSUE #113: Upgrade jwt-go version
golang/github.com/dgrijalva/jwt-go@3.2.0 package has a known vulnerability issue. 

_jwt-go before 4.0.0-preview1 allows attackers to bypass intended access restrictions in situations with []string{} for m["aud"] (which is allowed by the specification). Because the type assertion fails, "" is the value of aud. This is a security problem if the JWT token is presented to a service hat lacks its own audience check._ 

Link for more info https://ossindex.sonatype.org/vulnerability/c16fb56d-9de6-4065-9fca-d2b4cfb13020?component-type=golang&component-name=github.com%2Fdgrijalva%2Fjwt-go&utm_source=nancy-client&utm_medium=integration&utm_content=1.0.10

pyrooka:
This PR replaces the JWT package to a fork to prevent some possible vulnerabilities.

Closes #113


padamstx (approved):
LGTM

PR #116, opened 2021-05-21
feat: check for empty string in ParseDate

pyrooka:
This PR adds empty string check to the `ParseDate` function. It's the same as the `strfmt` packages has in the `ParseDateTime` function. Tests have been updated.


padamstx (approved):
LGTM

padamstx:
@pyrooka I did a squash&merge and decided to use a "fix" commit message instead of "feat".  I was thinking about that some more prior to merging and just thought the minor change was more of a "fix" than a "new feature" :).
Thanks for jumping on this one so quickly.

PR #118, opened 2021-05-28
fix: allow user to set "Host" request header

padamstx:
The golang net/http package's Request type will ignore
the "Host" header if set in the Request.Header field.
Instead, one must explicitly set this header in the Request.Host
field.

This commit includes changes to the RequestBuilder and
BaseService types to allow a user to set "Host" in the normal
ways that an SDK user can set request headers:
- as a custom header while constructing an IAM or CP4D authenticator
- as a default header set on a service instance
- as a custom header within an options model instance when
  invoking an operation
In each of these scenarios, if the user sets the "Host" header
then the resulting Request.Host field is set with this value.
If no "Host" header is specified by the user, then Request.Host
field would be set to the host:port specified in the request URL
(the default behavior).

### DIFF
--- a/v5/core/base_service.go
+++ b/v5/core/base_service.go
@@ -306,6 +306,15 @@ func (service *BaseService) Request(req *http.Request, result interface{}) (deta
 		for k, v := range service.DefaultHeaders {
 			req.Header.Add(k, strings.Join(v, ""))
 		}
+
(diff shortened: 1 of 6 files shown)


dpopp07 (approved):
Looks good! 👍

PR #119, opened 2021-05-28
feat: add `constructServiceURL` function

ajhynes7:
This function can be used by an API to construct a service URL from a parameterized URL.


v5/core/parameterized_url.go
```diff
@@ -0,0 +1,83 @@
...
+//
+func constructServiceURL(
+	parameterizedUrl string,
+	defaultUrlVariables map[string]string,
+	providedUrlVariables map[string]string,
+) (string, error) {
+
+	// If null was passed, we set the variables to an empty map.
+	// This results in all default variable values being used.
+	if providedUrlVariables == nil {
```
padamstx, line 45: Incidentally enough, it's ok to read from a nil map.  You just get back the zero value ("" for string) for any key that you try to retrieve.
Reference: https://blog.golang.org/maps

It's ok to account for a nil map the way you have done here; just thought I'd mention it.

v5/core/parameterized_url.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	// Verify the provided variable names.
+	for providedName := range providedUrlVariables {
+		if _, ok := defaultUrlVariables[providedName]; !ok {
+			// Get all variable names (the keys of the default variables map).
+			acceptedNames := make([]string, len(defaultUrlVariables))
+
+			i := 0
+			for name := range defaultUrlVariables {
```
padamstx, line 56: I think it would be cleaner to move the creation of the "acceptedNames" slice before this for loop on line 50, then combine that for loop together with the second one on line 73.
IOW, compute the list of valid variable names ahead of time, then enter a loop where you are formatting the URL AND checking for invalid variable names in one step.

v5/core/parameterized_url.go
```diff
@@ -0,0 +1,83 @@
...
+				acceptedNames[i] = name
```
padamstx, line 57:
```suggestion
				acceptedNames = append(acceptedNames, name)
```
When adding items to a slice in go, I think it is more idiomatic to use "append".  That way, you don't need to worry about the index variable, etc.
There is a slight performance hit in using append, but in this case it shouldn't be a concern since we're only dealing with a small number of variable names.

padamstx (changes requested):
Overall looks good, but I suggested a few improvements.

v5/core/parameterized_url.go
ajhynes7: I've changed the code to use `append` now.

I don't think I can combine the two loops because the first is iterating over the provided variables, and the second is iterating over the default variables. The first loop helps the user by catching if they misspelled a variable name. If we just looped over the default variables, the default value would be used instead of the intended provided value, so the user could get a URL that they didn't want.

I kept the creation of the `acceptedNames` inside the loop so it's only invoked if a provided variable name is invalid. Otherwise, we don't need to compute it.

padamstx: I see what you're saying about not being able to combine the two loops.   I didn't realize before that you were looping over entries in the two different maps :)

padamstx (approved):
LGTM

PR #120, opened 2021-06-21
fix: make the get token method exported

pyrooka:
This PR renames the `getToken()` methods in the IAM authenticator and in the CP4D authenticator to `GetToken()`.


padamstx (approved):
LGTM

PR #121, opened 2021-07-05
chore: switch to a more actively maintained JWT package

pyrooka:
Not so long ago, we replaced (18d04ad2f6e4fa32386898c39a4580eb4bca7910) the [dgrijalva/jwt-go](https://github.com/dgrijalva/jwt-go) package, because it was not maintained and there were a few unpatched security vulnerabilities in the code.
An official community fork has been created [golang-jwt/jwt](https://github.com/golang-jwt/jwt) and also the readme in the original repo (marked as archive) recommends to use that.

### DIFF
--- a/v5/core/jwt_utils.go
+++ b/v5/core/jwt_utils.go
@@ -19,7 +19,7 @@ import (
 	"fmt"
 	"strings"
 
-	jwt "github.com/form3tech-oss/jwt-go"
+	"github.com/golang-jwt/jwt"
 )
 
 // coreJWTClaims are the fields within a JWT's "claims" segment that we're interested in.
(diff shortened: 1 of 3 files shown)


padamstx:
@pyrooka Thank you for being proactive about this dependency.  However, I think we need to re-think how we use the "jwt" package and assess whether or not we really need a separate package for it.   Specifically, this new package apparently only officially supports go 1.15 and 1.16, whereas our Go core and SDKs support go 1.12+.

padamstx (approved):
I think the changes in this PR look fine as they are, but I also think we should consider removing our dependency on the jwt package altogether.

pyrooka:
Good idea! I will make that change in a separate PR and close this as soon as that will be merged in. (I think merging this PR is unnecessary.)

pyrooka:
Closing this in favor of #122

PR #122, opened 2021-07-08
chore: reimplement a JWT function to drop the dependency

pyrooka:
This PR implements the `decodeSegment()` function from the JWT package that we have been using and also completely removes the JWT dependency from the project.

Ref: #121

### DIFF
--- a/v5/core/jwt_utils.go
+++ b/v5/core/jwt_utils.go
@@ -15,11 +15,10 @@ package core
 // limitations under the License.
 
 import (
+	"encoding/base64"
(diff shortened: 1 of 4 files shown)


padamstx (approved):
Looks great!
Prior to merging this PR, I think it would be good to test the Platform Services Go SDK with this new go core version (you can use the replace directive in the go sdk's go.mod file to use a local copy of the go core).  Simply run a couple of integration tests to make sure we're exercising the IAM authenticator which will use this new function.

PR #126, opened 2021-07-27
fix: error message used by CR Authenticator

padamstx:
This commit changes the error message used by the
ComputeResourceAuthenticator to report a validation
error when neither of the `IAMProfileName` or `IAMProfileID`
properties are specified by the user.
The error message is now:
`At least one of IAMProfileName or IAMProfileID must be specified.`


dpopp07 (approved):
Looks good! 👍

PR #127, opened 2021-08-02
fix: remove VSI use-case from ComputeResourceAuthenticator

padamstx:
This commit removes the aspects of the new ComputeResourceAuthenticator
related to the VPC/VSI use-case.  We're doing this due to some changes
in the design of that use-case.
With these changes, the ComputeResourceAuthenticator will support
only the IKS file-based use-case.

Note that this PR most definitely introduces breaking changes, but I'm taking a calculated
risk that no Go core users have actually started using the ComputeResourceAuthenticator,
which I think is a risk that's worth taking :)
Otherwise, we'd need to deliver these changes in a new major version.


v5/core/cr_authenticator.go
```diff
@@ -101,40 +96,87 @@ type ComputeResourceAuthenticator struct {
...
-	iamGrantTypeCRToken    = "urn:ibm:params:oauth:grant-type:cr-token" // #nosec G101
-	crtokenLifetime        = 300
+	iamGrantTypeCRToken = "urn:ibm:params:oauth:grant-type:cr-token" // #nosec G101
 )
 
 var craRequestTokenMutex sync.Mutex
 
-// NewComputeResourceAuthenticator constructs a new ComputeResourceAuthenticator instance with the supplied values and
-// invokes the ComputeResourceAuthenticator's Validate() method.
-func NewComputeResourceAuthenticator(crtokenFilename string, instanceMetadataServiceURL, iamProfileName string, iamProfileID string,
```
padamstx, line 115: As part of these changes, I decided to go ahead and remote this traditional ctor function in favor of a "builder" instead.

v5/core/cr_authenticator.go
```diff
@@ -101,40 +96,87 @@ type ComputeResourceAuthenticator struct {
...
-		URL:                        url,
-		ClientID:                   clientID,
-		ClientSecret:               clientSecret,
-		DisableSSLVerification:     disableSSLVerification,
-		Scope:                      scope,
-		Headers:                    headers,
-	}
+// ComputeResourceAuthenticatorBuilder is used to construct an instance of the ComputeResourceAuthenticator
+type ComputeResourceAuthenticatorBuilder struct {
+	ComputeResourceAuthenticator
```
padamstx, line 106: Note that the builder really just embeds the actual authenticator struct to hold all the properties.   Makes it pretty convenient in the various setter and Build() methods below.

v5/core/cr_authenticator.go
```diff
@@ -430,44 +470,8 @@ func (authenticator *ComputeResourceAuthenticator) RequestToken() (*IamTokenServ
 	return tokenResponse, nil
 }
 
-// retrieveCRToken retrieves the CR token for the current compute resource.
```
padamstx, line 433: The diffs starting here to the end of the file are a bit confusing (at least to me!), but if you view the whole file, you can see that things are simplified quite a bit.

pyrooka (approved):
Looks good to me!

dpopp07 (approved):
Looks good! 👍

padamstx:
Closing this PR as I'll have more changes to make re: the name of the authenticator, etc.

PR #128, opened 2021-08-03
chore: codescan modifications

christiancompton:
This PR addresses some new false positives picked up by our codescans.

### DIFF
--- a/v5/core/cr_authenticator.go
+++ b/v5/core/cr_authenticator.go
@@ -100,6 +100,7 @@ type ComputeResourceAuthenticator struct {
 }
 
 const (
+	/* #nosec G101 */
 	defaultCRTokenFilename = "/var/run/secrets/tokens/vault-token"
 	defaultImdsEndpoint    = "http://169.254.169.254"
 	imdsVersionDate        = "2021-07-15"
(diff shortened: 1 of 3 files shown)


v5/core/cr_authenticator.go
```diff
@@ -405,6 +406,7 @@ func (authenticator *ComputeResourceAuthenticator) RequestToken() (*IamTokenServ
 	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
 		buff := new(bytes.Buffer)
 		_, _ = buff.ReadFrom(resp.Body)
+		/* #nosec G104 */
```
christiancompton, line 409: Unsure about whether this needs to be fixed instead of ignored.  This is a Low vulnerability, G104 indicates `"Errors unhandled."`

v5/core/cr_authenticator.go
```diff
@@ -574,6 +577,7 @@ func (authenticator *ComputeResourceAuthenticator) retrieveCRTokenFromIMDS() (cr
...
```
christiancompton, line 580: See above.

padamstx:
@christiancompton It turns out that I'm in the process of updating this code with lots of changes due to some upstream design changes.   I'll incorporate your first change in my updated code, and the second change is no longer applicable as that code has been removed altogether.    I think you can just close out this PR.

PR #129, opened 2021-08-03
fix: refactor container authenticator with recent design changes

padamstx:
This PR contains several changes to the authenticator formerly known as `ComputeResourceAuthenticator`:
- it has been renamed to `ContainerAuthenticator`
- it now implements only the IKS file-based cr token use-case (the parts related to the VPC use-case were removed)
- the authenticator's traditional ctor function was re-designed into a "builder"
- documentation updated to reflect latest changes


Authentication.md
```diff
@@ -278,39 +278,12 @@ The IAM access token is added to each outbound request in the `Authorization` he
    Authorization: Bearer <IAM-access-token>
 ```
 
-### Compute Resource Token Retrieval Algorithm
```
padamstx, line 281: No need to go into any detail about how the cr token is retreived since it's just read from a file.

pyrooka (approved):
Looks good!

v5/core/container_authenticator_test.go
```diff
@@ -30,62 +29,64 @@ import (
...
-	craTestRefreshToken   string   = "Xj7Gle500MachEOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJ1c2VybmFtZSI6ImhlbGxvIiwicm9sZSI6InVzZXIiLCJwZXJtaXNzaW9ucyI6WyJhZG1pbmlzdHJhdG9yIiwiZGVwbG95bWVudF9hZG1pbiJdLCJzdWIiOiJoZWxsbyIsImlzcyI6IkpvaG4iLCJhdWQiOiJEU1giLCJ1aWQiOiI5OTkiLCJpYXQiOjE1NjAyNzcwNTEsImV4cCI6MTU2MDI4MTgxOSwianRpIjoiMDRkMjBiMjUtZWUyZC00MDBmLTg2MjMtOGNkODA3MGI1NDY4In0.cIodB4I6CCcX8vfIImz7Cytux3GpWyObt9Gkur5g1QI"
+	containerAuthTestLogLevel       LogLevel = LevelError
+	containerAuthMockCRTokenFile    string   = "../resources/cr-token.txt"
+	containerAuthMockIAMProfileName string   = "iam-user-123"
+	containerAuthMockIAMProfileID   string   = "iam-id-123"
+	containerAuthMockClientID       string   = "client-id-1"
+	containerAuthMockClientSecret   string   = "client-secret-1"
+	containerAuthTestCRToken1       string   = "cr-token-1"
```
pyrooka, line 38: One question: wouldn't be better to rename the `cr-token` to simply `token` or `container-token`?

dpopp07: This is a good question. I know we've already messed with the naming plenty 🙂 but it's probably worth considering now that we've dropped the "compute resource" nomenclature from the user-facing code. This might apply to the `CRTokenFilename` property as well.

You do explain in the documentation that the token is a "compute resource token", so I think it's fine as is, but worth considering the simpler naming.

dpopp07 (approved):
Looks good! I just left one comment jumping on Norbert's question to get a little discussion there, but the changes look good to me

v5/core/container_authenticator_test.go
padamstx: I wanted to distinguish between the "cr token" and the "access token", hence the use of "cr token".   Even though we're now calling this the ContainerAuthenticator, the "token" is still considered to be a "compute resource token" (and that name can also be used in the VPCInstanceAuthenticator when we add that in as well).
If you guys are ok with it, I'd like to keep the current nomenclature ("cr token", "CRTokenFilename", etc.).

pyrooka: That make sense so it's ok for me.

dpopp07: That sounds good to me too 👍

PR #130, opened 2021-08-04
chore: codescan modifications

christiancompton:
This PR address code scan false positives with the exception of G104 errors, which may no longer be needed.

### DIFF
--- a/v5/core/constants.go
+++ b/v5/core/constants.go
@@ -45,7 +45,7 @@ const (
 	PROPNAME_CLIENT_ID        = "CLIENT_ID"
 	PROPNAME_CLIENT_SECRET    = "CLIENT_SECRET"
 	PROPNAME_SCOPE            = "SCOPE"
-	PROPNAME_CRTOKEN_FILENAME = "CR_TOKEN_FILENAME"
+	PROPNAME_CRTOKEN_FILENAME = "CR_TOKEN_FILENAME" // #nosec G101
 	PROPNAME_IAM_PROFILE_NAME = "IAM_PROFILE_NAME"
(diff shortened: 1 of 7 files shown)


v5/core/constants.go
```diff
@@ -45,6 +45,7 @@ const (
 	PROPNAME_CLIENT_ID        = "CLIENT_ID"
 	PROPNAME_CLIENT_SECRET    = "CLIENT_SECRET"
 	PROPNAME_SCOPE            = "SCOPE"
+	/* #nosec G101 */
```
padamstx, line 48: I hate to ask this but would it also work to use an inline "//-style" command at the end of the affected line?
If so, I think I'd prefer to just be consistent with the way in which we annotate code to avoid the false positives.

v5/core/container_authenticator.go
```diff
@@ -92,6 +92,7 @@ type ContainerAuthenticator struct {
 }
 
 const (
+	/* #nosec G101 */
 	defaultCRTokenFilename = "/var/run/secrets/tokens/vault-token"
 	iamGrantTypeCRToken = "urn:ibm:params:oauth:grant-type:cr-token" // #nosec G101
```
padamstx, line 97: example of existing inline comment

padamstx (commented):
Just had a question about being consistent with the types of comments that we use for the annotations.   All of the affected lines of code are false positives so I agree on all those.

christiancompton:
@padamstx I updated the new results to use inline comments, and updated the previous above line entries to be consistent throughout.

padamstx (approved):
Thank you for indulging my OCD-level nitpickiness :)

PR #132, opened 2021-08-10
chore: codescan updates

christiancompton:
The [`nosec` annotations](https://github.com/securego/gosec#annotating-code) are not processed when the preceding content is followed by a comma.  I tried several permutations to see if I could get around this, but was unsuccessful.


padamstx (approved):
LGTM

PR #134, opened 2021-08-12
build: add gosec scan to Travis

pyrooka:
This PR adds a security scan ([gosec](https://github.com/securego/gosec)) to the Travis CI pipeline and updates the golangci-lint version.


Makefile
```diff
@@ -4,7 +4,7 @@ VDIR=v5
 
 COV_OPTS=-coverprofile=coverage.txt -covermode=atomic
 
-all: testcov lint tidy
+all: testcov lint scan-gosec tidy
```
padamstx, line 7: I don't think we should include scan-gosec in the "all" target, to be honest.
Let's leave it as a separate target and then we can add a `make scan-gosec` command to the script section.  Would that work?

padamstx (changes requested):
I think it would be good to keep the "scan-gosec" target/goal separate from the "all" target.

Makefile
pyrooka: Yep, that makes sense.

padamstx (approved):
LGTM

PR #135, opened 2021-08-19
build: build on go 1.12 and 1.16

padamstx:
This PR updates the go core build so that it is built on go 1.12 and 1.16.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,10 @@
 language: go
 
-dist: xenial
+dist: bionic
 
 go:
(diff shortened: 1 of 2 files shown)


padamstx:
Currently the 1.16 build fails in the scan-gosec step.  I can reproduce these errors in my local environment as I'm running go v1.16.6 on my Fedora laptop.   Will work on fixing these code-scanning problems and update the PR.

barrett-schonefeld (approved):
Looks good!

PR #137, opened 2021-08-31
fix: handle the error during gzip compression instead of panic

pyrooka:
This PR fixes/improves the error handling in the Gzip compression. Before this commit if an error occurred during the compression it caused a panic. With this change, in case of errors, the pipe will be closed and the error will be passed.


padamstx (approved):
LGTM!

PR #138, opened 2021-09-14
fix: recognize vendor-specific JSON mimetypes

padamstx:
This commit modifies the Go core response-handling code
so that it recognizes vendor-specific JSON-based mimetypes
(e.g. application/vnd.docker.distribution.manifest.v2+json)
as being equivalent to "application/json" when deciding how
to unmarshal an operation response body.
Prior to this change, the Go core would treat such a response
as an error.


pyrooka (approved):
LGTM!

dpopp07 (approved):
Looks good! 👍

PR #140, opened 2021-10-01
chore(build): bump node to 14 to support new semantic-release

padamstx:
This PR moves us up to node 14 to support the new v18 of semantic-release.


dpopp07 (approved):
Looks good! 👍

padamstx:
I've added an additional commit to bump up our min supported go version to 1.14.
This is needed because a few of our dependencies require a go version newer than 1.12 (our previous min version).

pyrooka (approved):
LGTM!

PR #146, opened 2021-11-19
feat(IamAuthenticator): support refresh token flow in IamAuthenticator

padamstx:
This commit updates the IamAuthenticator so that it supports
the "refresh token" auth flow.  In this scenario, the user constructs
an IamAuthenticator with a refresh token (and client id/secret) instead
of an apikey.  The authenticator will use the "POST /identity/token"
operation with grant_type "refresh_token" to obtain an IAM access token.

In this commit, I also added a new "builder" for the IamAuthenticator
as well, and deprecated the legacy ctor function although it is still supported.


v5/core/iam_authenticator.go
```diff
@@ -85,29 +92,98 @@ var iamNeedsRefreshMutex sync.Mutex
...
-	defaultIamTokenServerEndpoint = "https://iam.cloud.ibm.com"              // #nosec G101
-	iamGrantTypeApiKey            = "urn:ibm:params:oauth:grant-type:apikey" // #nosec G101
+	defaultIamTokenServerEndpoint = "https://iam.cloud.ibm.com" // #nosec G101
+	iamauthOperationPathGetToken  = "/identity/token"
+	iamauthGrantTypeApiKey        = "urn:ibm:params:oauth:grant-type:apikey" // #nosec G101
+	iamauthGrantTypeRefreshToken  = "refresh_token"                          // #nosec G101
 )
 
-// NewIamAuthenticator constructs a new IamAuthenticator instance.
-func NewIamAuthenticator(apikey string, url string, clientId string, clientSecret string,
```
padamstx, line 93: This legacy ctor was simply moved down a bit.

v5/core/iam_authenticator.go
```diff
@@ -85,29 +88,98 @@ var iamNeedsRefreshMutex sync.Mutex
...
 		return nil, err
 	}
 
-	return authenticator, nil
+	return &builder.IamAuthenticator, nil
+}
+
+// NewIamAuthenticator constructs a new IamAuthenticator instance.
+// Deprecated - use the IamAuthenticatorBuilder instead.
+func NewIamAuthenticator(apikey string, url string, clientId string, clientSecret string,
```
pyrooka, line 171:
```suggestion
func NewIamAuthenticator(apiKey string, url string, clientId string, clientSecret string,
```

v5/core/iam_authenticator_test.go
```diff
@@ -28,13 +30,221 @@ import (
...
+	iamauthTestAccessToken2 string = "3yJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJ1c2VybmFtZSI6ImhlbGxvIiwicm9sZSI6InVzZXIiLCJwZXJtaXNzaW9ucyI6WyJhZG1pbmlzdHJhdG9yIiwiZGVwbG95bWVudF9hZG1pbiJdLCJzdWIiOiJoZWxsbyIsImlzcyI6IkpvaG4iLCJhdWQiOiJEU1giLCJ1aWQiOiI5OTkiLCJpYXQiOjE1NjAyNzcwNTEsImV4cCI6MTU2MDI4MTgxOSwianRpIjoiMDRkMjBiMjUtZWUyZC00MDBmLTg2MjMtOGNkODA3MGI1NDY4In0.cIodB4I6CCcX8vfIImz7Cytux3GpWyObt9Gkur5g1QI"
+	iamauthTestRefreshToken string = "Xj7Gle500MachEOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJ1c2VybmFtZSI6ImhlbGxvIiwicm9sZSI6InVzZXIiLCJwZXJtaXNzaW9ucyI6WyJhZG1pbmlzdHJhdG9yIiwiZGVwbG95bWVudF9hZG1pbiJdLCJzdWIiOiJoZWxsbyIsImlzcyI6IkpvaG4iLCJhdWQiOiJEU1giLCJ1aWQiOiI5OTkiLCJpYXQiOjE1NjAyNzcwNTEsImV4cCI6MTU2MDI4MTgxOSwianRpIjoiMDRkMjBiMjUtZWUyZC00MDBmLTg2MjMtOGNkODA3MGI1NDY4In0.cIodB4I6CCcX8vfIImz7Cytux3GpWyObt9Gkur5g1QI"
```
pyrooka, line 43:
```suggestion
	iamAuthTestLogLevel     LogLevel = LevelError
	iamAuthMockApiKey                = "mock-apikey"
	iamAuthMockRefreshToken          = "mock-refresh-token"
	iamAuthMockClientID              = "bx"
	iamAuthMockClientSecret          = "bx"
	iamAuthMockURL                   = "https://mock.iam.com"
	iamAuthMockScope                 = "scope1,scope2"

	iamAuthTestAccessToken1 string = "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJ1c2VybmFtZSI6ImhlbGxvIiwicm9sZSI6InVzZXIiLCJwZXJtaXNzaW9ucyI6WyJhZG1pbmlzdHJhdG9yIiwiZGVwbG95bWVudF9hZG1pbiJdLCJzdWIiOiJoZWxsbyIsImlzcyI6IkpvaG4iLCJhdWQiOiJEU1giLCJ1aWQiOiI5OTkiLCJpYXQiOjE1NjAyNzcwNTEsImV4cCI6MTU2MDI4MTgxOSwianRpIjoiMDRkMjBiMjUtZWUyZC00MDBmLTg2MjMtOGNkODA3MGI1NDY4In0.cIodB4I6CCcX8vfIImz7Cytux3GpWyObt9Gkur5g1QI"
	iamAuthTestAccessToken2 string = "3yJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJ1c2VybmFtZSI6ImhlbGxvIiwicm9sZSI6InVzZXIiLCJwZXJtaXNzaW9ucyI6WyJhZG1pbmlzdHJhdG9yIiwiZGVwbG95bWVudF9hZG1pbiJdLCJzdWIiOiJoZWxsbyIsImlzcyI6IkpvaG4iLCJhdWQiOiJEU1giLCJ1aWQiOiI5OTkiLCJpYXQiOjE1NjAyNzcwNTEsImV4cCI6MTU2MDI4MTgxOSwianRpIjoiMDRkMjBiMjUtZWUyZC00MDBmLTg2MjMtOGNkODA3MGI1NDY4In0.cIodB4I6CCcX8vfIImz7Cytux3GpWyObt9Gkur5g1QI"
	iamAuthTestRefreshToken string = "Xj7Gle500MachEOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJ1c2VybmFtZSI6ImhlbGxvIiwicm9sZSI6InVzZXIiLCJwZXJtaXNzaW9ucyI6WyJhZG1pbmlzdHJhdG9yIiwiZGVwbG95bWVudF9hZG1pbiJdLCJzdWIiOiJoZWxsbyIsImlzcyI6IkpvaG4iLCJhdWQiOiJEU1giLCJ1aWQiOiI5OTkiLCJpYXQiOjE1NjAyNzcwNTEsImV4cCI6MTU2MDI4MTgxOSwianRpIjoiMDRkMjBiMjUtZWUyZC00MDBmLTg2MjMtOGNkODA3MGI1NDY4In0.cIodB4I6CCcX8vfIImz7Cytux3GpWyObt9Gkur5g1QI"
```
Same as I mentioned above.

pyrooka (approved):
It looks good, just a few variable naming suggestions.

v5/core/iam_authenticator.go
padamstx: done

v5/core/iam_authenticator.go
```diff
@@ -85,29 +88,98 @@ var iamNeedsRefreshMutex sync.Mutex
 
 const (
 	// The default (prod) IAM token server base endpoint address.
-	defaultIamTokenServerEndpoint = "https://iam.cloud.ibm.com"              // #nosec G101
-	iamGrantTypeApiKey            = "urn:ibm:params:oauth:grant-type:apikey" // #nosec G101
+	defaultIamTokenServerEndpoint = "https://iam.cloud.ibm.com" // #nosec G101
+	iamauthOperationPathGetToken  = "/identity/token"
+	iamauthGrantTypeApiKey        = "urn:ibm:params:oauth:grant-type:apikey" // #nosec G101
+	iamauthGrantTypeRefreshToken  = "refresh_token"                          // #nosec G101
```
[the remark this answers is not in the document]
padamstx: done

v5/core/iam_authenticator_test.go
padamstx: done

v5/core/iam_authenticator.go
```diff
@@ -165,23 +240,40 @@ func (authenticator *IamAuthenticator) setTokenData(tokenData *iamTokenData) {
...
+	if this.ApiKey != "" && HasBadFirstOrLastChar(this.ApiKey) {
 		return fmt.Errorf(ERRORMSG_PROP_INVALID, "ApiKey")
 	}
 
-	// Validate ClientId and ClientSecret.  They must both be specified togther or neither should be specified.
-	if this.ClientId == "" && this.ClientSecret == "" {
+	// Validate ClientId and ClientSecret.
+	// If RefreshToken is not specified, then both or neither should be specified.
+	// If RefreshToken is specified, then both must be specified.
+	if this.ClientId == "" && this.ClientSecret == "" && this.RefreshToken == "" {
```
dpopp07, line 276: Is it possible that the API key was created without either the client ID or secret set? I guess there's a default that's used in that case but would the refresh token path use the default as well?

dpopp07 (approved):
Looks good! Just have one question but I'll approve since it's probably something you already have an answer to

v5/core/iam_authenticator.go
padamstx: Great questions...
When obtaining an access token for an apikey, there are no "default" values for client id/secret that are used if they're not specified. In this scenario, you won't get back a valid refresh token.

In order to obtain a valid refresh token, you must specify "bx/bx" for the client id/secret values along with the apikey.
So, if this refresh token is then going to be used to construct the IAM Authenticator, the same client id/secret values must be used as well.   I didn't want to hardcode client id/secret to be "bx/bx" in the authenticator when a refresh token is being used because that seems a bit hackish.  I discussed this with Martin Smolny and we settled on "the client id/secret values must be the same values that were used to initially obtain the refresh token", even though in the IBM IAM implementation those values an only be "bx/bx".  However, in the future, this constraint might be relaxed to support additional values for those properties, but the "same as" constraint would still apply... presumably :)

PR #149, opened 2021-12-10
fix: avoid false positive gosec errors

padamstx:
This PR tweaks a few of the inline code annotations used to ignore false-positive gosec violations.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ before_install:
 
 install:
   - curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh| sh -s -- -b $(go env GOPATH)/bin v1.41.1
-  - curl -sfL https://raw.githubusercontent.com/securego/gosec/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v2.8.1
+  - curl -sfL https://raw.githubusercontent.com/securego/gosec/master/install.sh | sh -s -- -b $(go env GOPATH)/bin
 
 script:
   - make all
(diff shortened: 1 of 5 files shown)


v5/core/container_authenticator.go
```diff
@@ -404,7 +404,7 @@ func (authenticator *ContainerAuthenticator) RequestToken() (*IamTokenServerResp
 		// If the user told us to disable SSL verification, then do it now.
 		if authenticator.DisableSSLVerification {
 			transport := &http.Transport{
-				/* #nosec G402 */
+				// #nosec G402
```
padamstx, line 407: For some reason, the latest version of gosec (2.9.4, which was released yesterday and apparently is used by Brent's security scanning) prefers `// #nosec` over `/* #nosec */` for inline code annotations.

.travis.yml
```diff
@@ -19,7 +19,7 @@ before_install:
 
 install:
   - curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh| sh -s -- -b $(go env GOPATH)/bin v1.41.1
-  - curl -sfL https://raw.githubusercontent.com/securego/gosec/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v2.8.1
+  - curl -sfL https://raw.githubusercontent.com/securego/gosec/master/install.sh | sh -s -- -b $(go env GOPATH)/bin
```
padamstx, line 22: Removed the version # so we'll always download/install the latest version.  That way, we'll see these failures sooner rather than later.

rmkeezer (approved):
LGTM

Issue #150, opened 2022-01-05
IAM Validate() fails post Authenticate()


mkumatag:
sample program:

```ini
$ cat ibm-credentials.env
IBMCLOUD_AUTH_TYPE=iam
IBMCLOUD_APIKEY=xxxxxxxxxxxxxxxxxx
IBMCLOUD_AUTH_URL=https://iam.cloud.ibm.com
```

```golang
package main

import (
	"fmt"
	"github.com/IBM/go-sdk-core/v5/core"
	"github.com/davecgh/go-spew/spew"
	"net/http"
)

func main() {
	auth, err := core.GetAuthenticatorFromEnvironment("IBMCLOUD")
	if err != nil {
		fmt.Printf("errored core.GetAuthenticatorFromEnvironment: %+v", err)
	}
	req, err := http.NewRequest("GET", "http://example.com", nil)
	if err != nil {
		fmt.Printf("errored http.NewRequest: %+v", err)
	}
	err = auth.Validate()
	if err != nil {
		fmt.Printf("errored auth.Validate before Authenticate: %+v", err)
	}
	err = auth.Authenticate(req)
	if err != nil {
		fmt.Printf("errored auth.Authenticate: %+v", err)
	}
	spew.Dump(auth)
	err = auth.Validate()
	if err != nil {
		fmt.Printf("errored auth.Validate after Authenticate: %+v", err)
	}
}
```

output:
```shell
$ go run sample-iam.go
(*core.IamAuthenticator)(0xc00014eab0)({
 ApiKey: (string) (len=44) "xxxxxxxxxxxxxxxxxxxxxx",
 RefreshToken: (string) (len=13) "not_supported",
 URL: (string) (len=25) "https://iam.cloud.ibm.com",
 ClientId: (string) "",
 ClientSecret: (string) "",
 DisableSSLVerification: (bool) false,
 Scope: (string) "",
 Headers: (map[string]string) <nil>,
 Client: (*http.Client)(0xc000351dd0)({
  Transport: (http.RoundTripper) <nil>,
  CheckRedirect: (func(*http.Request, []*http.Request) error) <nil>,
  Jar: (http.CookieJar) <nil>,
  Timeout: (time.Duration) 30s
 }),
 tokenData: (*core.iamTokenData)(0xc0005a2240)({
  AccessToken: (string) (len=1443) "xxxxxxxxxxxxxxxxxxxxxx",
  RefreshToken: (string) (len=13) "not_supported",
[...]

PR #151, opened 2022-01-05
Ignore refresh token if not_supported

mkumatag:
Fixes: #150

### DIFF
--- a/v5/core/iam_authenticator.go
+++ b/v5/core/iam_authenticator.go
@@ -249,7 +249,8 @@ func (authenticator *IamAuthenticator) setTokenData(tokenData *iamTokenData) {
 	// This was recommended by the IAM team to avoid problems in the future
 	// if the token service is changed to invalidate an existing refresh token
 	// when a new one is generated and returned in the response.
-	if tokenData != nil {
+	// Ignore if the refresh token is not_supported.
+	if tokenData != nil && tokenData.RefreshToken != "not_supported" {
 		authenticator.RefreshToken = tokenData.RefreshToken
 	}
 }


v5/core/iam_authenticator.go
```diff
@@ -249,7 +249,8 @@ func (authenticator *IamAuthenticator) setTokenData(tokenData *iamTokenData) {
 	// This was recommended by the IAM team to avoid problems in the future
 	// if the token service is changed to invalidate an existing refresh token
 	// when a new one is generated and returned in the response.
-	if tokenData != nil {
+	// Ignore if the refresh token is not_supported.
+	if tokenData != nil && tokenData.RefreshToken != "not_supported" {
```
padamstx, line 253: According to the IAM team, we should not be relying on "not_supported" to be the value when an invalid refresh token is returned by the IAM get_token operation.
Besides, this check wouldn't help the case where a valid refresh token is returned.
All you really need to do is to avoid calling Validate() and you will avoid the problem

padamstx (commented):
I don't think we want to pull in this change...

v5/core/iam_authenticator.go
mkumatag: In that case, do you think we need to change Validate() logic somehow to accommodate the changes, maybe a private field called `refreshToken` which is used only within the module instead of overwriting user-supplied one?

I have just written a sample program for repro purpose, but this is very well possible if someone shares an authenticator object between multiple services then there is a possibility that the first service calls the Validate, Authenticate, and then the second service calls again Validate and Authenticate(the second one will fail because the first service already feed that refresh_token)

PR #152, opened 2022-01-18
test: improve test coverage

padamstx:
This commit modifies several tests to increase code
coverage slightly.
In addition, there were a few minor tweaks made to
the functional code to improve testability, plus one
really minor bug fix in the BaseService.SetUserAgent()
method.

### DIFF
--- a/v5/core/base_service.go
+++ b/v5/core/base_service.go
@@ -278,7 +278,7 @@ func (service *BaseService) buildUserAgent() string {
 // SetUserAgent sets the user agent value.
 func (service *BaseService) SetUserAgent(userAgentString string) {
 	if userAgentString == "" {
-		service.UserAgent = service.buildUserAgent()
+		userAgentString = service.buildUserAgent()
 	}
 	service.UserAgent = userAgentString
 }
--- a/v5/core/container_authenticator.go
+++ b/v5/core/container_authenticator.go
@@ -54,7 +54,8 @@ type ContainerAuthenticator struct {
 
 	// [optional] The IAM token server's base endpoint URL.
 	// Default value: "https://iam.cloud.ibm.com"
-	URL string
+	URL     string
+	urlInit sync.Once
 
 	// [optional] The ClientID and ClientSecret fields are used to form a "basic auth"
 	// Authorization header for interactions with the IAM token server.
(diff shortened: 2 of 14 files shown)


v5/core/base_service.go
```diff
@@ -278,7 +278,7 @@ func (service *BaseService) buildUserAgent() string {
 // SetUserAgent sets the user agent value.
 func (service *BaseService) SetUserAgent(userAgentString string) {
 	if userAgentString == "" {
-		service.UserAgent = service.buildUserAgent()
+		userAgentString = service.buildUserAgent()
```
padamstx, line 281: This code coverage work uncovered one minor bug!!!

v5/core/container_authenticator.go
```diff
@@ -54,7 +54,8 @@ type ContainerAuthenticator struct {
 
 	// [optional] The IAM token server's base endpoint URL.
 	// Default value: "https://iam.cloud.ibm.com"
-	URL string
+	URL     string
+	urlInit sync.Once
```
padamstx, line 58: I updated the container and iam authenticators so that they initialize the URL field the same way the vpc authenticator does for consistency, plus it improves testability just a bit.

v5/core/vpc_instance_authenticator.go
```diff
@@ -280,11 +280,6 @@ func (authenticator *VpcInstanceAuthenticator) invokeRequestTokenData() error {
 // and then (2) exchange that for an IAM access token.
 func (authenticator *VpcInstanceAuthenticator) RequestToken() (iamTokenResponse *IamTokenServerResponse, err error) {
 
-	// Use the default VPC base endpoint if user didn't specifiy the URL property.
```
padamstx, line 283: This check is actually not needed because of the way we initialize and use the URL field (through the url() method) within the `retrieveInstanceIdentityToken()` and `retrieveIamAccessToken()` methods.

v5/core/base_service.go
pyrooka: That's a nice catch!

v5/core/log_test.go
```diff
@@ -55,6 +61,7 @@ func TestSetLogLevel(t *testing.T) {
 	assert.True(t, l.IsLogLevelEnabled(LevelWarn))
 	assert.True(t, l.IsLogLevelEnabled(LevelInfo))
 	assert.True(t, l.IsLogLevelEnabled(LevelDebug))
+
```
pyrooka, line 64: There is an extra line here.
```suggestion
```

pyrooka (approved):
Looks good!

PR #154, opened 2022-01-31
fix: allow retries and disable ssl to co-exist

padamstx:
Previously, the "disable ssl" and "auto retries"
scenarios were mutually exclusive due to the way in
which the http client instances were managed.
The result was that you could not enable retries
and also disable ssl verification.

This commit fixes this problem by adding more precise
code to configure the "disable ssl" feature and
also reuses existing client instances where possible.


v5/core/base_service.go
```diff
@@ -348,25 +391,9 @@ func (service *BaseService) Request(req *http.Request, result interface{}) (deta
 		}
 	}
 
+	// Invoke the request, then check for errors during the invocation.
 	var httpResponse *http.Response
-
```
padamstx, line 352: While trapsing through the "retryablehttp" package source code to learn what I needed to fix the "disable ssl" problem, I discovered that we do not need to distinguish between retryable and non-retryable clients at this level.  This is because when retries are enabled, the BaseService.Client will hold a "shim" client instance whose "Do" method can be called just like the non-retryable scenario and the retry logic will get triggered as expected.

v5/core/base_service.go
```diff
@@ -581,22 +608,63 @@ func getErrorMessage(responseMap map[string]interface{}, statusCode int) string
 // configuration, and then set it on the service instance.
 // If maxRetries and/or maxRetryInterval are specified as 0, then default values
 // are used instead.
+//
```
padamstx, line 611: Thought it might be good to add these details since it took me a while to work through the various datastructures
and figure this out :)

v5/core/base_service.go
```diff
@@ -626,21 +694,6 @@ func NewRetryableHTTPClient() *retryablehttp.Client {
 	return client
 }
 
-// getRetryableHTTPClient returns the "retryable" Client hidden inside the specified http.Client instance
```
padamstx, line 629: didn't need this function here any more, but the retries-related tests rely on it, so I moved it there.

v5/core/base_service.go
```diff
@@ -348,25 +391,9 @@ func (service *BaseService) Request(req *http.Request, result interface{}) (deta
...
 
+	// Invoke the request, then check for errors during the invocation.
 	var httpResponse *http.Response
-
-	// Try to get the retryable Client hidden inside service.Client
-	retryableClient := getRetryableHTTPClient(service.Client)
-	if retryableClient != nil {
-		retryableRequest, retryableErr := retryablehttp.FromRequest(req)
-		if retryableErr != nil {
-			err = fmt.Errorf(ERRORMSG_CREATE_RETRYABLE_REQ, retryableErr.Error())
```
pyrooka, line 358: As I see this constant (`ERRORMSG_CREATE_RETRYABLE_REQ`) is not in use anymore, so it should be deleted, right?

pyrooka (approved):
Looks really good!

v5/core/base_service.go
padamstx: yes, good catch.  I'll delete that.

dpopp07 (approved):
Looks good! 👍 

Left one small comment to address before merging but will approve to avoid the re-review

v5/core/base_service.go
```diff
@@ -260,6 +272,37 @@ func (service *BaseService) IsSSLDisabled() bool {
 	return false
 }
 
+// getTransportForSSL() will return the http.Transport instance
```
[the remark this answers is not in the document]
padamstx: fixed it

PR #157, opened 2022-03-17
fix: retain http.Client config when retries are enabled

padamstx:
This commit includes changes to the BaseService functionality
so that enabling automatic retries will not cause the loss of
any existing http.Client instance that might be set on the
BaseService.
A user can now use BaseService.SetHTTPClient() to set
a custom client instance with or without retries being enabled.
Also, the BaseService.GetHTTPClient() method was introduced
which will return the correct http.Client instance regardless
of whether retries are enabled or not.

### DIFF
--- a/v5/core/base_service.go
+++ b/v5/core/base_service.go
@@ -227,28 +227,53 @@ func (service *BaseService) SetDefaultHeaders(headers http.Header) {
 	service.DefaultHeaders = headers
 }
 
-// SetHTTPClient updates the client handling the requests.
+// SetHTTPClient will set "client" as the http.Client instance to be used
+// to invoke individual HTTP requests.
+// If automatic retries are currently enabled on "service", then
+// "client" will be set as the embedded client instance within
+// the retryable client; otherwise "client" will be stored
+// directly on "service".
 func (service *BaseService) SetHTTPClient(client *http.Client) {
 	setMinimumTLSVersion(client)
-	service.Client = client
+
+	if isRetryableClient(service.Client) {
+		// If "service" is currently holding a retryable client,
+		// then set "client" as the embedded client used for individual requests.
+		tr := service.Client.Transport.(*retryablehttp.RoundTripper)
+		tr.Client.HTTPClient = client
+	} else {
+		// Otherwise, just hang "client" directly off the base service.
+		service.Client = client
+	}
 }
 
-// DisableSSLVerification skips SSL verification.
-// This function sets a new http.Client instance on the service
-// and configures it to bypass verification of server certificates
-// and host names, making the client susceptible to "man-in-the-middle"
-// attacks.  This should be used only for testing.
+// GetHTTPClient will return the http.Client instance used
+// to invoke individual HTTP requests.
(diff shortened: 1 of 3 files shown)


v5/core/base_service.go
```diff
@@ -273,41 +300,9 @@ func (service *BaseService) IsSSLDisabled() bool {
 	return false
 }
 
-// getClientTransportForSSL() will return the http.Transport instance
```
padamstx, line 276: no longer needed

v5/core/base_service.go
```diff
@@ -273,41 +300,9 @@ func (service *BaseService) IsSSLDisabled() bool {
...
-	}
-
-	return nil
-}
-
 // setMinimumTLSVersion sets the minimum TLS version required by the client to TLS v1.2
 func setMinimumTLSVersion(client *http.Client) {
-	tr := getClientTransportForSSL(client)
-	if tr != nil {
+	if tr, ok := client.Transport.(*http.Transport); tr != nil && ok {
```
padamstx, line 305: we will now only call this method on an actual (non-retryable) http.Client instance

v5/core/base_service.go
```diff
@@ -698,15 +706,33 @@ func (l *httpLogger) Printf(format string, inserts ...interface{}) {
...
 	client.Logger = &httpLogger{}
 	client.CheckRetry = IBMCloudSDKRetryPolicy
 	client.Backoff = IBMCloudSDKBackoffPolicy
 	client.ErrorHandler = retryablehttp.PassthroughErrorHandler
-	setMinimumTLSVersion(client.HTTPClient)
+
+	if httpClient != nil {
+		// If a non-nil http client was passed in, then let's use that
+		// as our embedded client used to invoke individual requests.
+		client.HTTPClient = httpClient
```
pyrooka, line 730: Shouldn't we set the minimum TLS version on the client that user provides?

pyrooka (approved):
I left a few comments, but looks good overall!

v5/core/base_service.go
```diff
@@ -227,28 +227,53 @@ func (service *BaseService) SetDefaultHeaders(headers http.Header) {
...
 
-// DisableSSLVerification skips SSL verification.
-// This function sets a new http.Client instance on the service
-// and configures it to bypass verification of server certificates
-// and host names, making the client susceptible to "man-in-the-middle"
-// attacks.  This should be used only for testing.
+// DisableSSLVerification will configure the service to
+// skip the verification of server certificates and hostnames.
+// This will make the client susceptible to "man-in-the-middle"
+// attacks.  This should be used only for testing or in secure
```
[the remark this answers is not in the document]
padamstx: picky picky :)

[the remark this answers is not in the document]
pyrooka: You should know me by now...😄

v5/core/base_service.go
pyrooka: Fine by me, just wanted to check with you.

dpopp07 (approved):
Looks good! 👍

PR #158, opened 2022-03-22
fix(IamAuthenticator): tweak Validate() method to be more lenient

padamstx:
This commit adds a bit of leniency to the IamAuthenticator.Validate() method.
We need to do this to ensure that a single authenticator instance can be shared
among multiple service instances.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "go.sum|package-lock.json|^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2022-03-17T20:39:01Z",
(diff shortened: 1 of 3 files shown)


pyrooka (approved):
Looks good!

mkumatag:
> This commit adds a bit of leniency to the IamAuthenticator.Validate() method.
> We need to do this to ensure that a single authenticator instance can be shared
> among multiple service instances.

finally! thanks for the fix @padamstx. IMO we should also demand IAM squad not return anything if no refresh token set, `not_supported` value doesn't make any sense to me.

PR #159, opened 2022-03-29
refactor: use sync.Once to initialize auth http clients

padamstx:
This commit modifies the Container, CP4D, and IAM
authenticators so that they are all consistent with the
VPC authenticator in the way in which they obtain an
HTTP client instance for invoking operations.


pyrooka (approved):
Looks good!

PR #161, opened 2022-04-25
build: remove Go 1.14 support

pyrooka:
This PR removes Go 1.14 from the project and
updates [deprecated](https://tip.golang.org/doc/go1.16#ioutil) package imports.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ language: go
 dist: bionic
 
 go:
-- '1.14.x'
 - '1.16.x'
 - '1.17.x'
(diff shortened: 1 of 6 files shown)


padamstx (commented):
Change look good, but you also need to update the README.md :)

padamstx (changes requested):
As we discussed on slack, let's back out the code changes that would prevent us from continuing to work with go 1.14 since the current code will work with go 1.14 thru 1.18.

padamstx (approved):
LGTM

PR #162, opened 2022-05-09
feat: add function GetQueryParamAsInt

padamstx:
This commit introduces the "GetQueryParamAsInt()"
utility function that will be used by generated
code to perform automatic pagination.
Similar to GetQueryParam, but also converts
the retrieved value to an int64.

### DIFF
--- a/v5/core/utils.go
+++ b/v5/core/utils.go
@@ -266,28 +266,49 @@ func SliceContains(slice []string, contains string) bool {
 	return false
 }
 
-// GetQueryParam() returns a pointer to the value of query parameter `param` from urlStr,
+// GetQueryParam returns a pointer to the value of query parameter `param` from urlStr,
 // or nil if not found.
-func GetQueryParam(urlStr *string, param string) (*string, error) {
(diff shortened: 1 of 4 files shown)


v5/core/utils.go
```diff
@@ -268,26 +268,47 @@ func SliceContains(slice []string, contains string) bool {
...
-		return nil, nil
+		return
+	}
+
+	value = &v
+
+	return
+}
+
+// GetQueryParamAsInt() returns a pointer to the value of query parameter `param` from urlStr
```
pyrooka, line 296: By default the documentation strings should only contain the name without parentheses, but you can leave it as is. :)
```suggestion
// GetQueryParamAsInt returns a pointer to the value of query parameter `param` from urlStr
```

pyrooka: As I see the `GetQueryParam` function above has the same "issue".

pyrooka (approved):
Just a really minor thing, looks good otherwise!

v5/core/utils.go
padamstx: done

PR #165, opened 2022-08-17
chore: replace deprecated `ioutil` package with `io` and `os`

pyrooka:
This commit replaces the `ioutil` package with `io` and `os`,
because it's became deprecated in Go 1.16. As of the release
of Go 1.19, our linter started to throw errors due to this old imports.

The first commit contains import changes while the second is the result of running `gofmt` on the project. I used Go 1.19, that's why there are a lot of changes in the documentations. From the release notes: ["As part of this change, gofmt now reformats doc comments to make their rendered meaning clearer."](https://go.dev/doc/go1.19#go-doc) The reformatted code looks good to me, but I'm okay if we'd like to not include it in this PR.


padamstx (approved):
LGTM!

PR #169, opened 2023-01-06
feat: create multi-part form as a streamed request body

padamstx:
This commit modifies the RequestBuilder.Build() method's handling of multi-part forms so that the request body is created in a streamed fashion, rather than making an entire copy of each mime part's contents.
This change will be especially useful when a multi-part form contains one or more parts containing large files.

Signed-off-by: Phil Adams <phil_adams@us.ibm.com>


pyrooka (approved):
LGTM!

PR #172, opened 2023-01-09
fix: use package.json to manage semrel and its deps

padamstx:


.travis.yml
```diff
@@ -28,16 +28,12 @@ script:
 
 before_deploy:
   - nvm install 14
-  - npm install -g npm@6.x
```
padamstx, line 31: node 14 includes npm v6, so this step not needed

.gitignore
```diff
@@ -19,6 +19,5 @@ coverage.txt
 /.vscode/
 
 # files produced by "npm install" commands during build
-package.json
```
padamstx, line 22: Just removed package.json since we want to now manage it with git

PR #173, opened 2023-01-10
feat(utils): add helper for creating a byte array pointer

dpopp07:
We have pointer-creation functions for a few other types. There is now a need to have a function like this for byte arrays within our generated Terraform code, so I've added it alongside the others.

Signed-off-by: Dustin Popp <dpopp07@gmail.com>


padamstx (approved):
LGTM

Issue #175, opened 2023-02-22
BaseService.Request invoked without result does not close http.Response


rafalbigaj:
The method `BaseService.Request` invoked without result (`nil`) does not close `http.Response` and prevents connection to be put back into idle state.

We observed significant memory usage or quick `MaxConnsPerHost` limit exhaustion, when `BaseService.Request` is called with `nil` result.

PR #176, opened 2023-02-22
fix: BaseService.Request invoked without result does not close http.Response

rafalbigaj:
The method `BaseService.Request` invoked without result (`nil`) does not close `http.Response` and prevents connection to be put back into idle state.

We observed significant memory usage or quick `MaxConnsPerHost` limit exhaustion, when `BaseService.Request` is called with `nil` result.

Issue: https://github.com/IBM/go-sdk-core/issues/175

### DIFF
--- a/v5/core/base_service.go
+++ b/v5/core/base_service.go
@@ -387,7 +387,7 @@ func (service *BaseService) Request(req *http.Request, result interface{}) (deta
 
 	// If debug is enabled, then dump the request.
 	if GetLogger().IsLogLevelEnabled(LevelDebug) {
-		buf, dumpErr := httputil.DumpRequestOut(req, req.Body != nil)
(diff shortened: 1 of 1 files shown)


dpopp07 (approved):
Thanks for the contribution, Rafal!

This looks good to me, but I'd @padamstx to take a look at it too before merging

padamstx (approved):
LGTM

rafalbigaj:
Thanks @padamstx! Is there a way to merge this fix to v5.10.x?

padamstx:
> Thanks @padamstx! Is there a way to merge this fix to v5.10.x?

If it is really needed there, then yes, I can work out a way to get it merged into the v5.10 release.   But... the current 5.12.x release should be compatible.  After I merge the PR into main (it will produce the 5.12.2 release), would you be able to try out that out to see if we can avoid applying the fix to 5.10.x?

PR #183, opened 2023-05-02
fix(ContainerAuthenticator): add sa-token as default CR token filename

padamstx:
This commit adds support for a second default CR token filename. If the user doesn't specify a CR token filename, the authenticator will first try `/var/run/secrets/tokens/vault-token`, and then `/var/run/secrets/tokens/sa-token` in that order.

### DIFF
--- a/core/container_authenticator.go
+++ b/core/container_authenticator.go
@@ -1,6 +1,6 @@
 package core
 
-// (C) Copyright IBM Corp. 2021.
+// (C) Copyright IBM Corp. 2021, 2023..
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


core/container_authenticator.go
```diff
@@ -479,25 +481,41 @@ func (authenticator *ContainerAuthenticator) RequestToken() (*IamTokenServerResp
...
-	crTokenFilename := authenticator.CRTokenFilename
-	if crTokenFilename == "" {
-		crTokenFilename = defaultCRTokenFilename
+	if authenticator.CRTokenFilename != "" {
+		// Use the file specified by the user.
+		crToken, err = authenticator.readFile(authenticator.CRTokenFilename)
+	} else {
+		// If the user didn't specify a filename, try our two defaults.
+		crToken, err = authenticator.readFile(defaultCRTokenFilename1)
+		if err != nil {
```
pyrooka, line 490: I think this is perfectly fine here, but a question came to my mind and can't let it go. We would like to try to read the 2nd file, even if the error is not a `FileNotFound` but - let's say - a permission issue, right? Really just asking... :)

pyrooka (approved):
LGTM!

core/container_authenticator.go
padamstx: Good question... yes, I thought about the various errors that might occur and whether we should treat a permissions issue differently than a "file doesn't exist at all" issue.
I concluded that we should try to keep this simple and treat any error while trying to read the file the same... i.e. we couldn't read it (whatever the reason) so move on to the next one to try.  I think it's extremely unlikely that we'd encounter a scenario where both default files (vault-token and sa-token) exist and we end up using "sa-token" by mistake because "vault-token" is not readable by the application.

dpopp07 (approved):
Looks good and tested on internal cluster 👍

PR #184, opened 2023-05-05
fix(deps): bump dependencies to recent versions

padamstx:
Recent work with the cli systest framework shows some golang warnings about retracted versions of the go-mongodb module, so I updated version #'s of a few go core's dependencies to bring us current, and the bump of the strfmt module pulled in a newer version of go-mongodb so we should not see the warnings any more.


dpopp07 (approved):
Looks good! 👍

PR #189, opened 2023-08-17
docs: update header to match TOC list

ricellis:
Update the header to match the name used on line 5 of the file.


dpopp07 (approved):
Good catch, thanks for the PR! 👍

PR #199, opened 2023-11-06
feat(errors): create new error types to carry more info

dpopp07:
~This is a commit with the prototype code for the new error message formats. This isn't final but it's ready for a more thorough review and using a PR seems like the best way to do it. I'll mark this as a draft and tests are still failing as I clean things up but please begin to leave feedback and comments!~

The PR is now complete and ready for review.


core/errors.go
```diff
@@ -0,0 +1,206 @@
...
+  "crypto/sha256"
+  "encoding/hex"
+  "fmt"
+  "github.com/ghodss/yaml"
+  "strings"
+)
+
+// Type definitions
+
+type Problem interface {
```
padamstx, line 27: I think it would be good add a brief comment explaining each of the types defined in this file.  That would help folks quickly understand where the different types might be used, etc.

core/errors.go
```diff
@@ -0,0 +1,206 @@
...
+  GetConsoleMessage() string
+  GetDebugMessage() string
+  GetID() string
+}
+
+type IBMError struct {
+  Summary string // required
```
padamstx, line 34: We probably should also add comments to each of the public fields in the various types.

padamstx (approved):
Overall this looks good.
Here are some things to consider:
1. Perhaps we could consolidate a few of the "discriminator" values.  If the discriminator value is only one of multiple values that are used to create a "hash" then it might not be important to have super granular discriminator values.  For example, in jwt_utils.go perhaps all three calls to coreSDKErrorf() could use "invalid-token" or "invalid-jwt" for the discriminator.
2. We might want to scrutinize the various discriminator values that we use (I have my info developer hat on while making this statement :joy:).  Perhaps "invalid-input" instead of "bad-char", etc.
3. We might also want to include as part of this work changes to our error message constants so that the actual messages are not capitalized, which I think is idiomatic for Go (e.g. "Error while turning the light on" should be "error while turning the light on" :smile:)

core/errors.go
dpopp07: That's a great idea, I'll do that tomorrow

core/errors.go
dpopp07: Agreed

dpopp07:
We discussed this in person but I'll leave responses to your feedback here so we have it in writing

> Perhaps we could consolidate a few of the "discriminator" values. If the discriminator value is only one of multiple values that are used to create a "hash" then it might not be important to have super granular discriminator values. For example, in jwt_utils.go perhaps all three calls to coreSDKErrorf() could use "invalid-token" or "invalid-jwt" for the discriminator.

The purpose of the discriminator is to differentiate between errors that would otherwise have the same hash. For example, in `jwt_utils.go`, two of the calls to `coreSDKErrorf()` are in the same function but happen for different reasons. The discriminator is there to ensure we can capture different scenarios with different hashes. Therefore, I think it often helps for them to be fairly granular. We could potentially revisit how granular we want the hashes to be, though.

> We might want to scrutinize the various discriminator values that we use (I have my info developer hat on while making this statement 😂). Perhaps "invalid-input" instead of "bad-char", etc.

Something I should have documented is that the discriminator is not ever meant to be seen by the user. It's a private field, so it won't be accessible and it won't appear in the YAML-ified log messages. It's just there as a convenience field for developers to "discriminate" between errors. So, the user-friendliness shouldn't be _too_ relevant. That said, I agree it's worth taking a look to see if any of them detract from the readability of the code itself.

> We might also want to include as part of this work changes to our error message constants so that the actual messages are not capitalized, which I think is idiomatic for Go (e.g. "Error while turning the light on" should be "error while turning the light on" 😄)

Good idea - I hadn't been aware of that but I agree, this would be an appropriate time to make such a change.

core/errors.go
```diff
@@ -0,0 +1,206 @@
...
+      sdkError.System = system
+      sdkError.Version = version
+      sdkError.Function = function
+      err = sdkError
+    } else if httpError, ok := err.(*HTTPError); ok {
+      httpError.System = system
+      httpError.Version = version
+      httpError.OperationID = operationID
+      err = httpError
+    }
```
pyrooka, line 168: You can use type switch here which would improve the readability a little - in my opinion. Maybe it's just my personal preference, I let you to decide. :)
```suggestion
	switch _err := err.(type) {
	case SDKError:
		_err.System = system
		_err.Version = version
		_err.Function = function
		err = _err
	case HTTPError:
		_err.System = system
		_err.Version = version
		_err.OperationID = operationID
		err = _err
	}
```

pyrooka (approved):
It looks pretty good, I really like the logic. Left some comments to consider, but none of them is game changer.
Before I forget: I know this is just a draft, but please run the [goimports](https://pkg.go.dev/golang.org/x/tools/cmd/goimports) tool on the project or at least on the new `errors.go` file to fix the order of the imports (3rd party to the bottom, separated by a new line) and the indentation ([tab instead of spaces](https://www.youtube.com/watch?v=SsoOG6ZeyUI) 😛 ).

core/authenticator.go
```diff
@@ -29,15 +30,37 @@ type Authenticator interface {
 type AuthenticationError struct {
```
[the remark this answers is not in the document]
dpopp07: That's a good idea, I'll do that

core/errors.go
```diff
@@ -0,0 +1,206 @@
...
+func coreSDKErrorf(err error, summary, discriminator, function string) *SDKError {
+  system, version := getSystemInfo()
+  return &SDKError{
+    IBMError: ibmErrorf(err, summary, system, version, discriminator),
+    Function: function,
+  }
+}
+
+// Create new instances of `SDKError`
+func SDKErrorf(err error, summary, system, version, discriminator, function string) *SDKError {
```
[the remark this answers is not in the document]
dpopp07: Good catch! I haven't thoroughly documented everything yet but I'll make sure to follow this pattern as I do

core/errors.go
dpopp07: That's a good call, I like it! I'll make that change

core/authenticator_factory.go
```diff
@@ -61,7 +61,12 @@ func GetAuthenticatorFromEnvironment(credentialKey string) (authenticator Authen
 	} else if strings.EqualFold(authType, AUTHTYPE_NOAUTH) {
 		authenticator, err = NewNoAuthAuthenticator()
 	} else {
-		err = fmt.Errorf(ERRORMSG_AUTHTYPE_UNKNOWN, authType)
+		err = coreSDKErrorf(
+			nil,
+			fmt.Sprintf(ERRORMSG_AUTHTYPE_UNKNOWN, authType),
+			"unknown-auth-type",
```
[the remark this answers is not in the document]
dpopp07: We could, but they are not necessarily meant to be reused and in fact, it doesn't matter if they are re-used or not. They are just meant to force uniqueness between error hashes that would otherwise be the same (same function, etc.)

core/authenticator_factory.go
```diff
@@ -61,7 +61,12 @@ func GetAuthenticatorFromEnvironment(credentialKey string) (authenticator Authen
...
+			"GetAuthenticatorFromEnvironment",
```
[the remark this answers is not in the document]
dpopp07: Oh, that's nifty! I'll play around with that, thanks for sharing

bartlomiej-malecki:
As potential consumer of this PR - i would like to extend HTTPError with 

```
type HTTPError struct {
...
	// StatusCode same as Response.StatusCode
	StatusCode int

	// Errors array of `errors` from Response
	// if Response contains `error` instead of `errors` it will be the only element of Errors
	Errors []Error
}
```

so i could do something like
```
	err = callClientFunction()
	if err != nil {
		var httpErr HTTPError
		if errors.As(err, &httpErr) {
			if httpErr.StatusCode == 404 {
				return
			}
			fmt.Println(httpErr.Errors)
		}
	}
```

vast majority of errors we have to handle is coming from:
https://github.com/IBM/go-sdk-core/blob/main/core/base_service.go#L457

assuming we do this call in some handler in our microservice and we want e.g. to fail request with proper array of errors from downstream services, that would help a lot (we can do all this using DetailedResponse, but if we had this in error that would be very convenient)

dpopp07:
> As potential consumer of this PR - i would like to extend HTTPError

Thank you for posting the feedback! The more eyes we can get on this change, the better. I'll discuss this proposed change with the team. That is helpful to be aware of.

> PS. please don not change current value returned by err.Error()
i bet there is a lot of existing code

Great call. It has been my goal to keep the existing error messages intact so that `err.Error()` would return a consistent string. I'll double check the code to make sure I didn't thoughtlessly change anything.

Thanks again for looking this over and providing feedback. Feel free to share any other thoughts you have.

dpopp07:
@padamstx @pyrooka @hudlow this is now ready for review. I need to add some tests and finalize a couple of details but the bulk of the logic should remain as it is now. I also might split some of the new logic into separate files because the `errors.go` file is a bit monolithic. But it should be okay to review as is.

dpopp07:
> we can do all this using DetailedResponse, but if we had this in error that would be very convenient

@bartlomiej-malecki unfortunately, this was a bit too complicated to make it into the initial MVP scope but I still think it's a great idea and I hope to deliver it as a follow-on soon. It should be doable in a compatible fashion.

core/error_utils.go
```diff
@@ -0,0 +1,188 @@
...
+		return strings.TrimPrefix(runtime.FuncForPC(pc).Name(), component+"/")
+	}
+
+	return ""
+}
+
+// sdkStackFrame is a convenience struct for formatting
+// frame data to be printed as YAML.
+type sdkStackFrame struct {
+	Function string `json:"function,omitempty"`
```
padamstx, line 70: Should we have `json` tags on these fields or `yaml` tags if we're going to be doing yaml serialization?

padamstx (approved):
Looks great.  I added a few ticky-tack comments to consider.

core/bearer_token_authenticator.go
```diff
@@ -43,7 +44,7 @@ func NewBearerTokenAuthenticator(bearerToken string) (*BearerTokenAuthenticator,
 // newBearerTokenAuthenticator : Constructs a new BearerTokenAuthenticator instance from a map.
 func newBearerTokenAuthenticatorFromMap(properties map[string]string) (*BearerTokenAuthenticator, error) {
 	if properties == nil {
-		return nil, fmt.Errorf(ERRORMSG_PROPS_MAP_NIL)
+		return nil, SDKErrorf(nil, ERRORMSG_PROPS_MAP_NIL, "missing_props", getComponentInfo)
```
pyrooka, line 47: Typo?
```suggestion
		return nil, SDKErrorf(nil, ERRORMSG_PROPS_MAP_NIL, "missing-props", getComponentInfo)
```

core/container_authenticator.go
```diff
@@ -204,7 +203,7 @@ func (authenticator *ContainerAuthenticator) client() *http.Client {
 // configuration properties.
 func newContainerAuthenticatorFromMap(properties map[string]string) (authenticator *ContainerAuthenticator, err error) {
 	if properties == nil {
-		return nil, fmt.Errorf(ERRORMSG_PROPS_MAP_NIL)
+		return nil, SDKErrorf(nil, ERRORMSG_PROPS_MAP_NIL, "missing_props", getComponentInfo)
```
pyrooka, line 206: Typo?
```suggestion
		return nil, SDKErrorf(nil, ERRORMSG_PROPS_MAP_NIL, "missing-props", getComponentInfo)
```

core/base_service.go
```diff
@@ -612,6 +649,39 @@ func getErrorMessage(responseMap map[string]interface{}, statusCode int) string
 	return http.StatusText(statusCode)
 }
 
+// getErrorCode: try to retrieve an error code from the decoded response body (map).
```
pyrooka, line 652: I think this is more godoc compliant, but feel free to leave it as-is.
```suggestion
// getErrorCode tries to retrieve an error code from the decoded response body (map).
```

pyrooka (approved):
Well...it's impressive and all looks really good! I left a few small comments to consider, but that's all.
Actually, I have a few questions:
1. Wouldn't be good (in long term) to create and use the discriminator strings as constants? E.g. having a `discriminator.go` so we can use them as `discriminator.ObjIsNil`.
2. Wouldn't it worth to pass the `err` objects to the "constructors"  like `SDKErrorf` every time, even if they are not used? I am thinking of future scenarios where we might need them... 😄 
3. I know this is the only initial work and the change is already huge, but I am pretty sure we want to have unit tests for the new functions in the future.

core/errors.go
```diff
@@ -0,0 +1,453 @@
...
+	GetID() string
+
+	// Error returns the message associated with a given problem and guarantees
+	// every instance of Problem also implements the native `error` interface.
+	Error() string
+}
+
+// IBMProblem holds the base set of fields that all problem types
+// should include. It is geared towards embedding in other
+// structs and it should not be used on its own (so it is not exported).
```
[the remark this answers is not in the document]
dpopp07: Good call! That comment was leftover from when it was private, but it needs to be public to use it in Terraform. I'll update the comment.

core/errors.go
```diff
@@ -0,0 +1,453 @@
...
+type IBMProblem struct {
+
+	// Summary is the informative, user-friendly message that describes
+	// the problem and what caused it.
+	Summary string `json:"summary" validate:"required"` // required
```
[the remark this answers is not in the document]
dpopp07: We don't, so you're right. These are also leftover from an earlier prototype. I'll remove them everywhere.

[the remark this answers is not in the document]
dpopp07: Agreed, I was adding these for my own edification while planning out the schemas but they are unnecessary and unhelpful - I'm removing them.

core/errors.go
```diff
@@ -0,0 +1,453 @@
...
+	// for the developer to identify the root cause of the issue.
+	GetDebugMessage() string
+
+	// GetID returns an identifier or code for a given problem. It is computed
+	// from the attributes of the problem, so that the same problems will always
+	// have the same ID, even when encountered by different users.
+	GetID() string
+
+	// Error returns the message associated with a given problem and guarantees
+	// every instance of Problem also implements the native `error` interface.
```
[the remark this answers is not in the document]
dpopp07: Haha I think that's fair! I'd rather be consistent with the rest of the core, so I'll replace them with double quotes.

core/error_utils.go
dpopp07: Good call. Actually, with the YAML package I ended up using, the keys are "lowercased" automatically when marshalling so we don't need them at all. I will remove them.

core/errors.go
```diff
@@ -0,0 +1,453 @@
...
+		stack:      stack,
+	}
+}
+
+// RepurposeSDKProblem provides a convenient way to take a problem from
+// another function in the same component and contextualize it to the current
+// function. Should only be used in public (exported) functions.
+func RepurposeSDKProblem(err error, discriminator string) error {
+	if err == nil {
+		return err
```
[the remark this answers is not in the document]
dpopp07: Fair enough - I'll make that change!

core/iam_authenticator.go
```diff
@@ -204,13 +203,13 @@ func NewIamAuthenticator(apiKey string, url string, clientId string, clientSecre
...
 
-	return authenticator, err
+	return authenticator, RepurposeSDKProblem(err, "auth-builder-fail")
 }
 
 // newIamAuthenticatorFromMap constructs a new IamAuthenticator instance from a map.
 func newIamAuthenticatorFromMap(properties map[string]string) (authenticator *IamAuthenticator, err error) {
 	if properties == nil {
-		return nil, fmt.Errorf(ERRORMSG_PROPS_MAP_NIL)
+		return nil, SDKErrorf(nil, ERRORMSG_PROPS_MAP_NIL, "missing_props", getComponentInfo)
```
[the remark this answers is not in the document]
dpopp07: Yes, thank you!

dpopp07:
@pyrooka good questions! I'll answer here:

> Wouldn't be good (in long term) to create and use the discriminator strings as constants? E.g. having a discriminator.go so we can use them as discriminator.ObjIsNil.

It's a good idea but I still want to avoid using constants for a couple of reasons. The biggest one is, we need to keep in mind how we are going to enumerate the possible errors within this package. Having the string value of each discriminator right in the error constructor makes statically computing the hash value for that created error much easier. I know that's not the best reason in the world to make code decisions but it's a very real problem before us. That, and the fact that many of the discriminators only have one usage and I'm not sure it's worth writing a whole new file for keeping track of them. Some values are used more than once, like in authenticators, but that doesn't actually matter that much. They shouldn't ever change once I commit them but even if one did or if I messed one up, consistency doesn't add any value beyond aesthetic.

> Wouldn't it worth to pass the err objects to the "constructors" like SDKErrorf every time, even if they are not used? I am thinking of future scenarios where we might need them... 😄

The errors are only meant to be passed to the constructors if they are coming from a separate IBM Cloud component, like an HTTP service or another SDK. They are used as the "caused by" error to chain them together. Even though we have to accept an `error` type, the argument should really should be a `Problem`. I prefer to keep that clear by only passing in the variable when we know we want to use it as a "caused by". If we need the errors in the future (the only thing we could ever really use from them is the message), we can always just pass them in without a compatibility issue (I think).

> I know this is the only initial work and the change is already huge, but I am pretty sure we want to have unit tests for the new functions in the future.

100% - not just for the future, but now 🙂 I am working on them for this PR.

pyrooka:
Thanks for the answer @dpopp07!

Re: the discriminator constants, it makes sense and I think you have already explained to me. I just wanted to make sure I understand your approach.

Re: passing the errors - okay now I see.

> 100% - not just for the future, but now 🙂 I am working on them for this PR.

Sorry, I might missed that but agree, we should have them before merging. I just didn't want to be pushy and say "LGTM, but can't approve without proper unit tests...". :)

PR #200, opened 2023-11-08
build: use `goimports` to format code and check during builds

pyrooka:
This commit adds a new step to the lint step in the builds, which
utilizes the `goimports` tool to check the code style and to report
an error if there is a discrepancy. Furthermore it adds a new `make`
target to automatically format all code in place.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,7 @@
 GO=go
 LINT=golangci-lint
 GOSEC=gosec
+FORMATTER=goimports
 
 COV_OPTS=-coverprofile=coverage.txt -covermode=atomic
 
@@ -15,9 +16,14 @@ test:
 
 lint:
 	${LINT} run --build-tags=all
+	${FORMATTER} -d core
+	if [[ -n `${FORMATTER} -d core` ]]; then exit 1; fi
 
 scan-gosec:
 	${GOSEC} ./...
+
+format:
+	${FORMATTER} -w core
 
 tidy:
 	${GO} mod tidy
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,7 @@ before_install:
 install:
   - curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh| sh -s -- -b $(go env GOPATH)/bin v1.51.2
   - curl -sfL https://raw.githubusercontent.com/securego/gosec/master/install.sh | sh -s -- -b $(go env GOPATH)/bin
+  - go install golang.org/x/tools/cmd/goimports@latest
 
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -15,9 +16,14 @@ test:
 
 lint:
 	${LINT} run --build-tags=all
+	${FORMATTER} -d core
+	if [[ -n `${FORMATTER} -d core` ]]; then exit 1; fi
```
padamstx, line 20: We don't really need to run this twice here, do we?
OTOH, perhaps this was a test to make sure I was paying attention :joy:

padamstx (commented):
Looks good with these caveats:
1. It looks like we're running goimports twice in the "lint" target of Makefile
2. I was surprised that there were no required changes to source files to comply with `goimports`.  I guess we were already in compliance?

Makefile
pyrooka: Yes, you've passed the test! 😂  
Jokes aside, I found that the easiest and most reliable way to print the diff emitted by the formatter (it prints nothing if the code is correct) and exit with code 1 if there is something to change. Unfortunately the formatter always exit with 0.

pyrooka: Okay, I've changed it to call it only once, let me know what you think! :)

Makefile
```diff
@@ -15,9 +16,13 @@ test:
 
 lint:
 	${LINT} run --build-tags=all
+	DIFF=$$(${FORMATTER} -d core); if [[ -n "$$DIFF" ]]; then printf "\n$$DIFF" && exit 1; fi
```
padamstx, line 19: Interesting, I'd never seen the use of $$ like this before (i.e. to capture the output of a command).  I was aware of `$(<command>)`, but not `$$(<command>)`.  In fact, I can't really even find where that is documented in the bash man page - lol.  I know it works ok on Linux, though, because I pulled down your branch and tried it :)

padamstx: Ah, now I think I know what's going on...  I think the extra $'s are needed because these commands are in a Makefile, rather than a bash script, amirite?  Do I win a prize? :joy:

padamstx (approved):
LGTM!

Makefile
pyrooka: Haha, of course! This is yours: 🥇
 
I wanted to avoid embedding a "complicated" shell script into the Makefile or having a separate script, but overall I think this oneliner is not that bad.

padamstx: agreed, this is fine.  I just missed the fact that it was in a make file.

dpopp07 (approved):
Looks good! 👍

PR #206, opened 2024-02-07
build: bump Go versions

pyrooka:
This commit adds the latest Go version to the builds (`1.22`)
and removes the oldest one (`1.19`), so now the minimum
version is `1.20`.


dpopp07 (approved):
👍

PR #209, opened 2024-02-23
fix: adjust IAM token expiration time

padamstx:
This commit changes the IAM, Container and VPC Instance authenticators slightly so that an IAM access token will be viewed as "expired" when the current time is within 10 seconds of the official expiration time. IOW, we'll expire the access token 10 secs earlier than the IAM server-computed expiration time.
We're doing this to avoid a scenario where
an IBM Cloud service receives a request along
with an "almost expired" access token and then uses that token to perform downstream requests in a
somewhat longer-running transaction and then the
access token expires while that transaction is
still active.

### DIFF
--- a/core/container_authenticator_test.go
+++ b/core/container_authenticator_test.go
@@ -2,7 +2,7 @@
 
 package core
 
-// (C) Copyright IBM Corp. 2021.
+// (C) Copyright IBM Corp. 2021, 2024.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -375,7 +375,7 @@ func TestContainerAuthGetTokenSuccess(t *testing.T) {
 
 	// Call GetToken() again and verify that we get the cached value.
 	// Note: we'll Set Scope so that if the IAM operation is actually called again,
-	// we'll receive the second access token.  We don't want the IAM operation called again yet.
+	// we'll receive the second access token. We don't want the IAM operation called again yet.
 	auth.Scope = "send-second-token"
 	accessToken, err = auth.GetToken()
 	assert.Nil(t, err)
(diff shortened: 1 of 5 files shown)


core/container_authenticator_test.go
```diff
@@ -392,6 +392,56 @@ func TestContainerAuthGetTokenSuccess(t *testing.T) {
...
+	auth.Scope = "send-second-token"
+	accessToken, err = auth.GetToken()
+	assert.Nil(t, err)
+	assert.Equal(t, containerAuthTestAccessToken1, accessToken)
+
+	// Force expiration and verify that GetToken() fetched the second access token.
+	// We'll set expiration to be current-time + <iamExpirationWindow> (10 secs),
+	// to test the scenario where we should refresh the token when we are within 10 secs
+	// of expiration.
+	auth.getTokenData().Expiration = GetCurrentTime() + iamExpirationWindow
```
padamstx, line 435: the `iamExpirationWindow` variable is defined in the IAM authenticator code and re-used by the testcases.

core/iam_authenticator.go
```diff
@@ -533,23 +537,28 @@ func newIamTokenData(tokenResponse *IamTokenServerResponse) (*iamTokenData, erro
 }
 
 // isTokenValid: returns true iff the IamTokenData instance represents a valid (non-expired) access token.
-func (this *iamTokenData) isTokenValid() bool {
-	if this.AccessToken != "" && GetCurrentTime() < this.Expiration {
+func (td *iamTokenData) isTokenValid() bool {
+	// We'll use "exp - 10" so that we'll treat an otherwise valid (unexpired) access token
```
padamstx, line 541: This is the key to the fix... the isTokenValid() function belonging to the "iamTokenData" struct.

core/iam_authenticator_test.go
```diff
@@ -388,7 +388,7 @@ func TestIamGetTokenSuccess(t *testing.T) {
 	assert.Nil(t, authenticator.synchronizedRequestToken())
 
 	// Force expiration and verify that we got the second access token.
-	authenticator.getTokenData().Expiration = GetCurrentTime() - 3600
+	authenticator.getTokenData().Expiration = GetCurrentTime() - 1
```
padamstx, line 391: No need to subtract 3600 (1 hour) from the current time to cause the "expiration" to occur.

core/iam_authenticator_test.go
```diff
@@ -442,6 +504,10 @@ func TestIamGetTokenSuccessRT(t *testing.T) {
 				"expiration": %d,
 				"refresh_token": "%s"
 			}`, iamAuthTestAccessToken2, expiration, newRefreshToken)
+			username, password, ok := r.BasicAuth()
```
padamstx, line 507: since this test is actually relying on the refresh token functionality, I just added a check to the mock server-side code to make sure we're passing in bx:bx in the basic auth header

core/iam_authenticator.go
```diff
@@ -292,7 +296,7 @@ func (authenticator *IamAuthenticator) setTokenData(tokenData *iamTokenData) {
 //
 // Ensures that the ApiKey and RefreshToken properties are mutually exclusive,
 // and that the ClientId and ClientSecret properties are mutually inclusive.
-func (this *IamAuthenticator) Validate() error {
+func (authenticator *IamAuthenticator) Validate() error {
```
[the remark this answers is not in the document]
pyrooka: I am wondering why I haven't seen this before. I am using golangci-lint as the lint tool in VS Code and it includes `staticcheck`which is the default one.
A funny side note: I started exploring the code base of a bigger Go project and in the [contributing doc](https://github.com/jesseduffield/lazygit/blob/master/CONTRIBUTING.md#go) they say:

> There is no shame in using self as a receiver name in a struct method. In fact we encourage it

core/container_authenticator_test.go
```diff
@@ -392,6 +392,56 @@ func TestContainerAuthGetTokenSuccess(t *testing.T) {
...
+
+	// Verify that the access token was returned by GetToken() and also
+	// stored in the authenticator's tokenData field as well.
+	assert.NotNil(t, auth.getTokenData())
+	assert.Equal(t, containerAuthTestAccessToken1, accessToken)
+	assert.Equal(t, containerAuthTestAccessToken1, auth.getTokenData().AccessToken)
+
+	// Call GetToken() again and verify that we get the cached value.
+	// Note: we'll Set Scope so that if the IAM operation is actually called again,
+	// we'll receive the second access token.  We don't want the IAM operation called again yet.
```
pyrooka, line 425:
```suggestion
	// we'll receive the second access token. We don't want the IAM operation called again yet.
```
Sorry!

pyrooka (approved):
Looks great and I really like the thorough unit tests!

core/container_authenticator_test.go
padamstx: I fixed that one plus one other instance of that same issue that you missed, young man :joy:

dpopp07 (approved):
Looks good! 👍

PR #211, opened 2024-03-14
fix(deps): upgrade yaml package to avoid vulnerability

dpopp07:
The changes also moves the dependency to the Go
module-compatible version of the package.


pyrooka (approved):
Looks good!

PR #212, opened 2024-03-19
fix(errors): support error checking methods for new problem structures

dpopp07:
Adds previously returned errors to the "Unwrap" chain for checking against in methods like "errors.Is()" and adds an implementation of "Is" to Problem types to enable equality checking against problems.

Also, adds all of the previously returned errors back into the problem structures to maintain full compatibility with downstream code.

### DIFF
--- a/core/base_service.go
+++ b/core/base_service.go
@@ -409,11 +409,10 @@ func (service *BaseService) Request(req *http.Request, result interface{}) (deta
 	httpResponse, err = service.Client.Do(req)
 
 	if err != nil {
-		errMsg := err.Error()
 		if strings.Contains(err.Error(), SSL_CERTIFICATION_ERROR) {
-			errMsg = ERRORMSG_SSL_VERIFICATION_FAILED + "\n" + errMsg
+			err = fmt.Errorf(ERRORMSG_SSL_VERIFICATION_FAILED + "\n" + err.Error())
 		}
-		err = SDKErrorf(nil, errMsg, "no-connection-made", getComponentInfo())
+		err = SDKErrorf(err, "", "no-connection-made", getComponentInfo())
 		return
 	}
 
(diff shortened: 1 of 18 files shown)


core/base_service.go
```diff
@@ -409,11 +409,10 @@ func (service *BaseService) Request(req *http.Request, result interface{}) (deta
 	httpResponse, err = service.Client.Do(req)
 
 	if err != nil {
-		errMsg := err.Error()
 		if strings.Contains(err.Error(), SSL_CERTIFICATION_ERROR) {
-			errMsg = ERRORMSG_SSL_VERIFICATION_FAILED + "\n" + errMsg
+			err = fmt.Errorf(ERRORMSG_SSL_VERIFICATION_FAILED + "\n" + err.Error())
 		}
-		err = SDKErrorf(nil, errMsg, "no-connection-made", getComponentInfo())
+		err = SDKErrorf(err, "", "no-connection-made", getComponentInfo())
```
dpopp07, line 415: @pyrooka Norbert - I want to publicly acknowledge that "you told me so" with respect to saving the previously returned errors in case we ever needed them! 😂 

Turns out, we do need them - so this PR is adding everything back in just like you recommended in my last PR 😛

core/ibm_problem.go
```diff
@@ -109,12 +116,21 @@ func (e *IBMProblem) GetCausedBy() Problem {
 // it does not include the error instance the method is called on - that is
 // looked at separately by the "errors" package in functions like "As".
 func (e *IBMProblem) Unwrap() []error {
+	var errs []error
```
dpopp07, line 119: The changes to this method are the key component of this PR. These changes re-enable compatibility with methods like `errors.Is` by including the previously returned errors in the list of "unwrapped" errors.

The changes to add `Is` implementations to the problem types are purely extra - they allow for comparisons against problems that go beyond being the same instance. They enable equality for the same problem scenario (i.e. same ID) which is what the new types are all about.

core/base_service.go
padamstx: Some developers never learn... :joy:

padamstx (approved):
Looks great!

core/base_service.go
pyrooka: Haha, thanks for mentioning me, appreciate it! But I hope you didn't do this to protect yourself from future complications that might come up, so you can point at me and say "Hey, that was Norbert's idea, not mine..." 😂

pyrooka (approved):
Looks good!

PR #213, opened 2024-03-21
fix(errors): restore matching against all original errors

dpopp07:
I missed an error origination point that needed to be converted to the new format. Also, this is maybe overly cautious but I added back in the original error instances as native caused by errors for all instances. Before, I intentionally left alone places that held an error origination from `fmt.Errorf()` because I don't know that users can really match against those, but I figured we might as well be as compatibility-friendly as we can be, since it doesn't cost us anything.

### DIFF
--- a/core/request_builder.go
+++ b/core/request_builder.go
@@ -105,7 +105,7 @@ func (requestBuilder *RequestBuilder) WithContext(ctx context.Context) *RequestB
 // invalid URL string (e.g. ":<badscheme>").
 func (requestBuilder *RequestBuilder) ConstructHTTPURL(serviceURL string, pathSegments []string, pathParameters []string) (*RequestBuilder, error) {
 	if serviceURL == "" {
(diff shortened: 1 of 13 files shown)


core/request_builder.go
```diff
@@ -143,7 +143,8 @@ func (requestBuilder *RequestBuilder) ConstructHTTPURL(serviceURL string, pathSe
 // The resulting request URL: "https://myservice.cloud.ibm.com/resource/res-123-456-789-abc/type/type-1"
 func (requestBuilder *RequestBuilder) ResolveRequestURL(serviceURL string, path string, pathParams map[string]string) (*RequestBuilder, error) {
 	if serviceURL == "" {
-		return requestBuilder, fmt.Errorf(ERRORMSG_SERVICE_URL_MISSING)
+		err := fmt.Errorf(ERRORMSG_SERVICE_URL_MISSING)
+		return requestBuilder, SDKErrorf(err, "", "service-url-missing", getComponentInfo())
```
dpopp07, line 147: This was the real issue that needed to be fixed

padamstx (approved):
LGTM

PR #214, opened 2024-04-08
fix(errors): wrap errors from external authenticators to use new system

dpopp07:
External authenticators may not use our error types. We previously assumed that errors coming from the `Authenticate` method would be either an `AuthenticationError` or `SDKProblem` and performed a type-cast that could cause a panic. This makes the type checking more explicit, and in the case of an unexpected error type, wraps the error in a new SDKProblem object and returns it to keep our error types consistent.


padamstx (approved):
LGTM

PR #216, opened 2024-04-12
feat: send user-agent header with auth token requests

padamstx:
This commit updates our various request-based authenticators so that the User-Agent header is included with each outbound token request. The value of the User-Agent header will be of the form "ibm-go-sdk-core/<authenticator-type>-<core-version> <os-info>".


pyrooka (approved):
LGTM!

dpopp07 (approved):
Looks good! 👍

PR #217, opened 2024-05-06
fix: use correct error variable in external auth case

dpopp07:
A typo resulted in a bug that prevents errors from being created in the event of external authenticators. This fixes the issue.


padamstx (approved):
LGTM

PR #218, opened 2024-05-07
fix(errors): flatten sdk problem chains to reduce hash complexity

dpopp07:
Previously, any SDKProblem instance could be used as the "caused by" for another SDKProblem instance and its ID would become a part of the hash. This meant that problems created between service SDKs and the Go SDK Core would have a lot of complexity - the error scenarios tracked would be unnecessarily granular and numerous.

This change updates the system to maintain one problem scenario per *context* (where a context is either HTTP, SDK, or Terraform) rather than per every component within a given context.

It also provides for keeping knowledge of problem scenarios originating in the core for easier debugging and easier attaching of HTTP errors to service SDK problems.


pyrooka (approved):
Looks good!

padamstx (approved):
LGTM

PR #219, opened 2024-05-15
fix(request_builder): encode unresolved path string prior to path parameter insertion

padamstx:
This commit modifies RequestBuilder.ResolveRequestURL() slightly so that the passed-in path string (unresolved, potentially containing special characters and path parameter references) is URL-encoded prior to replacing path param references with their corresponding encoded path param values.  This ensures that the URL string passed to http.NewRequest() by RequestBuilder.Build() will be properly encoded (escaped) such that all special characters appearing in path segments are converted to their %nn equivalents, but yet any escaped characters within the encoded path param values are NOT decoded.  This avoids a quirk in the url.Parse() method that is invoked by the http.NewRequest() method.

This change is needed in order to properly support a scenario where:
1. a path parameter value contains one or more embedded slashes and is therefore escaped AND
2. the unresolved path string ALSO contains other special characters needing to be escaped

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ test:
 
 lint:
 	${LINT} run --build-tags=all
-	DIFF=$$(${FORMATTER} -d core); if [[ -n "$$DIFF" ]]; then printf "\n$$DIFF" && exit 1; fi
+	DIFF=$$(${FORMATTER} -d core); if [ -n "$$DIFF" ]; then printf "\n$$DIFF\n" && exit 1; fi
 
 scan-gosec:
 	${GOSEC} ./...
--- a/core/sdk_problem_test.go
+++ b/core/sdk_problem_test.go
@@ -239,7 +239,7 @@ func TestSDKErrorfNoSummary(t *testing.T) {
 }
 
 func TestSDKErrorfDoesntUseSDKCausedBy(t *testing.T) {
-	sdkProb := getPopulatedSDKProblem();
+	sdkProb := getPopulatedSDKProblem()
 	newSDKProb := SDKErrorf(sdkProb, "", "", NewProblemComponent("a", "b"))
 	assert.Nil(t, newSDKProb.causedBy)
 	assert.NotNil(t, newSDKProb.nativeCausedBy)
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -19,7 +19,7 @@ test:
 
 lint:
 	${LINT} run --build-tags=all
-	DIFF=$$(${FORMATTER} -d core); if [[ -n "$$DIFF" ]]; then printf "\n$$DIFF" && exit 1; fi
+	DIFF=$$(${FORMATTER} -d core); if [ -n "$$DIFF" ]; then printf "\n$$DIFF\n" && exit 1; fi
```
padamstx, line 22: It turns out that this statement was always failing on Travis but yet it did not cause the build to stop due to a difference between "bash" and "sh" (the default shell on a Travis build agent is "sh").

Example build hiccup:
https://app.travis-ci.com/github/IBM/go-sdk-core/jobs/621470281

```
$ make lint
golangci-lint run --build-tags=all
DIFF=$(goimports -d core); if [[ -n "$DIFF" ]]; then printf "\n$DIFF" && exit 1; fi
/bin/sh: 1: [[: not found
The command "make lint" exited with 0.
```

padamstx: I verified that any formatting type differences will now cause the build to fail and "make" will return a non-zero exit status.

core/sdk_problem_test.go
```diff
@@ -239,7 +239,7 @@ func TestSDKErrorfNoSummary(t *testing.T) {
 }
 
 func TestSDKErrorfDoesntUseSDKCausedBy(t *testing.T) {
-	sdkProb := getPopulatedSDKProblem();
+	sdkProb := getPopulatedSDKProblem()
```
padamstx, line 242: change by "make format"

core/sdk_problem_utils.go
```diff
@@ -108,13 +108,13 @@ func formatFrames(pcs []uintptr, componentName string) []sdkStackFrame {
 }
 
 type sparseSDKProblem struct {
-	ID string
+	ID       string
```
padamstx, line 111: changes by "make format"

dpopp07 (approved):
Looks good! 👍

Makefile
pyrooka: Wow, good catch! I think I was the one who wrote that line and... I never thought about the shell difference to be honest - since we use Ubuntu images that usually use bash by default.

pyrooka (approved):
LGTM!

PR #221, opened 2024-06-25
fix: tidy modules

pyrooka:
In a recent change, the `go.sum` file was not updated correctly, so this
commit contains the result of running the `go mod tidy` command.


dpopp07 (approved):
Looks good! 👍

PR #223, opened 2024-07-26
build: add detect-secrets stage to build

padamstx:
This PR re-organizes the Travis build a bit in order to add a "detect-secrets" stage to the build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,11 +2,6 @@ language: go
 
 dist: jammy
 
-go:
-- '1.20.x'
-- '1.21.x'
-- '1.22.x'
-
 notifications:
   email: false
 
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -2,43 +2,54 @@ language: go
...
+    - go: '1.22.x'
+
+    - name: Detect-Secrets
+      language: python
+      python: 3.12
+      install:
+        - pip install --upgrade "git+https://github.com/ibm/detect-secrets.git@master#egg=detect-secrets"
+      script:
+        - detect-secrets scan --update .secrets.baseline
+        - detect-secrets -v audit --report --fail-on-unaudited --fail-on-live --fail-on-audited-real .secrets.baseline
```
dpopp07, line 43: Just out of curiosity, why not just add these steps to the above "script" section, where the tests are run, etc. ?

This looks good and looks clean overall, I'm just curious why you took this approach.

dpopp07 (approved):
Looks good to me! Had a question about the approach but I'll approve to avoid re-review (though @pyrooka might want to chime in on Monday 🙂)

.travis.yml
padamstx: Mainly so that everything is run in parallel.  Under the "Build-Test" stage, there are four different jobs that each run in parallel:
image

padamstx: And to carry this a little bit further (farther? :joy:), when we merge the PR into main, we'll see the "Semantic-Release" stage get run also:
image

.travis.yml
```diff
@@ -2,43 +2,54 @@ language: go
...
+      language: node_js
+      node_js: 18
+      before_install:
+        - pip install --user bump-my-version
+        - node --version
+        - npm --version
+      install:
+        - npm install
+      script:
+        - echo npm run semantic-release
```
padamstx, line 55: Oops, I just realized I need to fix this :joy:

padamstx: now fixed

pyrooka (approved):
Nothing I can complain about (😂), this looks good and clean indeed.

.travis.yml
dpopp07: Ah okay, cool!

PR #225, opened 2024-07-31
build: enable gosec linter in golangci-lint

padamstx:
This commit removes the use of the gosec standalone tool and enables the gosec linter supported by golangci-lint. I also fixed a bunch of gosec errors that were identified by the gosec linter.


pyrooka (approved):
LGTM!

PR #227, opened 2024-08-16
fix(logging): improve go core's debug logging

padamstx:
This commit adds some debug logging in various
components, and also modifies some tests so that
it's easy to enable debug logging when running
individual sets of tests.

### DIFF
--- a/core/base_service.go
+++ b/core/base_service.go
@@ -1,6 +1,6 @@
 package core
 
-// (C) Copyright IBM Corp. 2019, 2022.
+// (C) Copyright IBM Corp. 2019, 2024.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -18,7 +18,6 @@ import (
 	"bytes"
 	"context"
 	"crypto/tls"
-	"crypto/x509"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -131,6 +130,8 @@ func (service *BaseService) Clone() *BaseService {
 
 // ConfigureService updates the service with external configuration values.
 func (service *BaseService) ConfigureService(serviceName string) error {
+	GetLogger().Debug("Configuring BaseService instance with service name: %s\n", serviceName)
+
 	// Try to load service properties from external config.
 	serviceProps, err := getServiceProperties(serviceName)
 	if err != nil {
@@ -144,7 +145,7 @@ func (service *BaseService) ConfigureService(serviceName string) error {
 
 		// URL
 		if url, ok := serviceProps[PROPNAME_SVC_URL]; ok && url != "" {
-			err := service.SetURL(url)
+			err := service.SetServiceURL(url)
 			if err != nil {
 				err = RepurposeSDKProblem(err, "set-url-fail")
 				return err
@@ -221,6 +222,7 @@ func (service *BaseService) SetServiceURL(url string) error {
 	}
 
 	service.Options.URL = url
+	GetLogger().Debug("Set service URL: %s\n", url)
 	return nil
 }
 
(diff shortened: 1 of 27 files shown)


core/config_utils_test.go
```diff
@@ -25,6 +25,11 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+var (
+	// To enable debug logging during test execution, set this to "LevelDebug"
+	configUtilsTestLogLevel LogLevel = LevelDebug
```
pyrooka, line 30: Shouldn't this be configured to error level?

core/base_service.go
```diff
@@ -826,6 +834,16 @@ var (
...
+
+	// A regular expression to match the error returned by net/http when a
+	// request header or value is invalid. This error isn't typed
+	// specifically so we resort to matching on the error string.
+	invalidHeaderErrorRe = regexp.MustCompile(`invalid header`)
+
+	// A regular expression to match the error returned by net/http when the
+	// TLS certificate is not trusted. This error isn't typed
+	// specifically so we resort to matching on the error string.
+	notTrustedErrorRe = regexp.MustCompile(`certificate is not trusted`)
```
pyrooka, line 846: You've probably made a more thorough investigation than me - just right now, but according to the following code snippet from Go's source code, we might need to add a Windows specific regex pattern to make sure this package is truly cross-platform.
```go
macosErr:   "x509: “valid.testing.golang.invalid” certificate is not trusted",
windowsErr: "x509: certificate signed by unknown authority",
```
[source](https://github.com/golang/go/blob/aeac0b6cbfb42bc9c9301913a191bb09454d316a/src/crypto/x509/platform_test.go#L171)

pyrooka (approved):
I left 2 minor suggestions, otherwise it looks pretty good!

core/base_service.go
padamstx: When I originally wrote the `IBMCloudSDKRetryPolicy()` function, I essentially stole most of the code from the `go-retryablehttp.ErrorPropagatedRetryPolicy`, located [here](https://github.com/hashicorp/go-retryablehttp/blob/40b0cad1633fd521cee5884724fcf03d039aaf3f/client.go#L486).

It looks like you referenced a section of a testcase.  Are we sure that on windows we do in fact get a different error message?

pyrooka: I resolve this thread as we discussed it on Slack. Phil suggested to use
```go
notTrustedErrorRe = regexp.MustCompile(`certificate is not trusted|certificate is signed by unknown authority`)
```
and I think that's a good idea. It should cover most if not all situations.

core/config_utils_test.go
padamstx: yes... yes it should :joy:

padamstx: fixed it

core/base_service.go
padamstx: I updated the regex to include both substrings.

core/base_service.go
```diff
@@ -289,6 +291,7 @@ func (service *BaseService) DisableSSLVerification() {
 		// Disable server ssl cert & hostname verification.
 		tr.TLSClientConfig.InsecureSkipVerify = true // #nosec G402
 	}
+	GetLogger().Debug("Disabled SSL verification in HTTP client")
```
dpopp07, line 294: This message seems helpful but I don't think I remember seeing this in the other cores (or at least, not in Node). Perhaps we should add it there as well? (Maybe I missed it)

dpopp07 (approved):
Looks good! 👍

core/base_service.go
padamstx: yeah, it turns out that we have no `disableSslVerification()` in the node core's BaseService class.

dpopp07: Oh, interesting. I think there is in the authenticators in the Node core. Odd that we aren't consistent there. Perhaps we should be, but that can be a follow-on effort.

PR #232, opened 2024-12-05
fix: wrap full validation errors instead of just their message

pyrooka:
This PR contains a simple adjustment to error handling of the struct field validation.
Previously only the string representation of the error was included in the returned
`SDKError`, but an upcoming change to the CLI plugins requires the actual error
object, because it contains valuable information.


padamstx (approved):
LGTM
This is clearly an example of "addition by subraction" :joy:

dpopp07 (approved):
Looks good! 👍

PR #236, opened 2025-01-07
fix: avoid runtime panic when the `errors` field is `nil`

pyrooka:
This PR adds an extra step to the error message processing chain,
so that the length of the `errors` slice is checked before trying to
get the first error from it.


pyrooka:
This PR will be held back until Travis starts functioning again.

pyrooka:
@padamstx This is a trust game. There is no Travis to help you out. Only you and me. 😂

core/base_service.go
```diff
@@ -627,7 +627,11 @@ func getErrorMessage(responseMap map[string]interface{}, statusCode int) string
 		var errors Errors
 		responseBuffer, _ := json.Marshal(responseMap)
 		if err := json.Unmarshal(responseBuffer, &errors); err == nil {
-			return errors.Errors[0].Message
+			if len(errors.Errors) > 0 {
```
padamstx, line 630: Wow, how did we exist for so long with this bug in place?  LOL
This must be the first time that a service returned an error response with an explicit `null` value for "errors".

Please check the other cores to make sure we don't have a similar bug there.

padamstx (approved):
LGTM

padamstx:
@pyrooka After I got GH actions working on the main branch (and published a release), I rebased your PR branch on top of the latest main branch, then force-pushed it.  Since everything is now green, I'll merge it.

PR #243, opened 2025-02-13
build: bump Go versions, update dependencies and `golangci-lint`

pyrooka:
This commit removes Go 1.21 and adds Go 1.24. Also it updates the project dependencies and `golangci-lint` to the current latest version.


padamstx (approved):
LGTM

PR #245, opened 2025-03-06
fix(build): bump crypto dependency to avoid CVE

padamstx:


pyrooka (approved):
LGTM! One question though: the `.travis.yml` didn't get updated, which is fine since it's not in use anymore. Shouldn't we just remove it to avoid confusions? We can always get it back from the git history if we ever need it again. What do you think?

padamstx:
> Shouldn't we just remove it to avoid confusions?

Yes, we should.  I will make it so :)

PR #250, opened 2025-05-16
feat(auth): add support for MCSP V2 authentication

padamstx:
This commit adds the MCSPV2Authenticator implementation. This authenticator will invoke the MCSP V2 POST /api/2.0/{scopeCollectionType}/{scopeId}/apikeys/token operation to obtain an access token for a user-supplied apikey.


pyrooka (approved):
Everything looks pretty good!

PR #251, opened 2025-06-09
fix(BaseService): allow User-Agent to be set via service-level headers

padamstx:
This commit modifies the Go core's BaseService.Request() method slightly so that when setting service-level custom headers on the outbound request, it will now replace any existing header with the service-level header value, rather than add the service-level header value to the existing value that might already exist.
The net effect of this is that a User-Agent header set as a service-level default header will now override a User-Agent header set by an SDK project through its "common.GetSdkHeaders()" function.

### DIFF
--- a/core/base_service_test.go
+++ b/core/base_service_test.go
@@ -2,7 +2,7 @@
 
 package core
 
-// (C) Copyright IBM Corp. 2019, 2024.
+// (C) Copyright IBM Corp. 2019, 2025.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


pyrooka (approved):
Looks good! What do you think about adding a simple test case (or extending an existing one) to cover this change in the logic? It's totally up to you! :)

core/base_service_test.go
```diff
@@ -1032,12 +1032,15 @@ func TestClient(t *testing.T) {
 	assert.ObjectsAreEqual(mockClient, service.Client)
 }
 
-func TestRequestForDefaultUserAgent(t *testing.T) {
+func TestRequestUserAgentDefault(t *testing.T) {
```
padamstx, line 1035: we now have 3 similarly-structured testcases that test the various scenarios (default, operation-level, service-level).

PR #258, opened 2025-10-16
build: migrate golangci-lint to v2

diatrcz:
Migrate golangci-lint to v2 from v1.


.golangci.yaml
```diff
@@ -1,6 +1,26 @@
...
+      - third_party$
+      - builtin$
+      - examples$
+formatters:
+  exclusions:
+    generated: lax
+    paths:
+      - third_party$
+      - builtin$
+      - examples$
```
pyrooka: Let's try removing the entire `formatters` block to pass all files to the formatters. Using the defaults should be the best on the long run, but first, let's see if it works. 😅

pyrooka (commented):
I left a few comments to consider. Haven't tested all of them so expect lint failures, but we should keep things default and as strict as possible. Let me know if you have any questions!

These comments/suggestions also apply to https://github.com/IBM/platform-services-go-sdk/pull/411

pyrooka (approved):
LGTM!

PR #265, opened 2026-03-25
build: exclude some gosec rules from golangci-lint

diatrcz:
`golangci-lint` needed to be upgraded for the builds to run with Go version `1.26`. Because of this some newly added rules had to be disabled.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -10,10 +10,9 @@ linters:
       - std-error-handling
   settings:
     gosec:
-      includes:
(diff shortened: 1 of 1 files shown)


pyrooka (commented):
Is `- all` valid here? According to the config, we might need to use the `exclude` field to prevent some rules being checked: https://golangci-lint.run/docs/linters/configuration/#gosec

pyrooka (approved):
LGTM!

PR #270, opened 2026-05-28
fix(deps): bump sys dependency to avoid CVE

Andris28:
In this commit we bump the sys
dependency to avoid CVE opened 
by scanners.


diatrcz (approved):
lgtm!

PR #271, opened 2026-05-29
Update utils.go

jacek-midura:
issue https://github.ibm.com/arf/planning-sdk-squad/issues/4393 
Too much data is redacted if the last path param in the URL is on `redactedKeywords` list.


pyrooka (changes requested):
First off, thanks for the contribution. Before approving this I'd like to ask 2 things:
1. Add sign-off to your commits in the branch to make the DCO check pass
2. Add new test cases to cover the change you made. Currently multiline strings are not covered as far as I remember.

PR #272, opened 2026-06-01
fix(cli): correct inline golangci-lint error

Andris28:
In this commit we fix an issue where
the golangci-lint would fail.


pyrooka (approved):
LGTM!

PR #278, opened 2026-07-09
fix: fix disabled golangci-lint rules

diatrcz:
When updating `golangci-lint` from v1 to v2 a lot of new linters were added which resulted in a lot of new lint errors. Those rules got disabled. This PR enables them and fixes the raised linter errors.


pyrooka (approved):
Looks good!

